PR #1, opened 2021-02-19
Adds Ecto Adapter

warmwaffles:


warmwaffles:
I've been pouring over https://github.com/elixir-ecto/ecto_sql and https://github.com/elixir-ecto/ecto to figure out what callbacks need to be implemented.

I think the closes implementation we should probably look to would be the MySQL implementation. Rationale for that, `myxql` has to deal with similar issues sqlite has where we are missing the `returning` ability for insertion.

I don't think it is so important to support that feature.

warmwaffles:
I think the worst part about the current set of adapters that exist within `ecto_sql` is that they are really hard to test. There are a bunch of private utility methods that are buried deep within the code and make it difficult to test and verify things.

I'm going to come up with something a little more manageable soon.

PR #2, opened 2021-02-20
First steps towards DB Connection

dmitriid:
I've split this into many small commits :D

~I'm now stuck with `handle_execute` because Sqlite3 expects a SQL statement, but `prepare` returns a reference. Haven't looked into this yet.~ I finally figured it out in #3 🙄 

I'm following myxql somewhat closely, but currently only implementing the most barebone functionality. Also see inline #TODO comments

I've tested what I have so far manually.

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -18,4 +18,29 @@ defmodule Exqlite.Connection do
     :db
   ]
 
+  @impl true
+  def connect(opts) do
+    path = Keyword.get(opts, :path)
+
+    case path do
+      nil ->
+        {:error,
+         %Error{
+           message:
+             "You must provide a :path to the database. Example: connect(path: \"./\") or connect(path: :memory)"
+         }}
+
+      _ ->
+        do_connect(path)
+    end
+  end
+
+  @impl true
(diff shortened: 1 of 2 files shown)


warmwaffles:
If you go take a look at #1 you'll see that I follow `myxql` closely, but I also pulled bits and pieces out of `sqlite_ecto2`. That PR is more just a big glob of changes as I try to understand and work through some of the decisions made in `ecto_sql`

lib/exqlite/connection.ex
```diff
@@ -0,0 +1,140 @@
...
+  """
+  @impl true
+  def handle_begin(opts, state) do
+    # TODO: track nested transactions.
+    #       sqlite doesn't nest transactions with BEGIN... COMMIT...,
+    #       use the SAVEPOINT and RELEASE commands.
+
+    # TODO: handle/track SAVEPOINT
+    case Keyword.get(opts, :mode, :deferred) do
+      :immediate -> handle_transaction(:begin, "BEGIN IMMEDIATE TRANSACTION", state)
```
warmwaffles, line 85: cool I completely forgot about these modes for the transaction 👍🏻

warmwaffles (approved):

PR #3, opened 2021-02-20
Db Connection implementation: handle_execute + prepare + bind

dmitriid:
This builds on #2.

Relevant commits: 

- 1242c1e `handle_execute` implementation. Includes prepare and bind
- 404ce4d and 99f4f09 are minor code fixes to handle this

Note: everything is as barebones as possible, just to check that everything is working. So, no funny majick around bind, for example :) See inline comments for TODOs

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -1,6 +1,5 @@
 defmodule Exqlite.Connection do
-
-  @doc """
+  @moduledoc """
   This module imlements connection details as defined in DBProtocol.
 
   Notes:
@@ -13,6 +12,7 @@ defmodule Exqlite.Connection do
   alias Exqlite.Sqlite3
   alias Exqlite.Error
   alias Exqlite.Result
+  alias Exqlite.Query
 
   defstruct [
     :db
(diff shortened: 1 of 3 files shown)


warmwaffles:
@dmitriid could you do a rebase or update your branch based on what is merged in?

dmitriid:
@warmwaffles Did a quick rebase.

I've seen your commits into #1 but haven't looked at them in detail 👍 It looks like we're doing more-or-less the same thing. You're just doing it faster :D

warmwaffles:
@dmitriid I've just been cobbling together stuff and refactoring. There is just so much extra cruft in all of this.

I'm going to go the route you are taking and make these smaller chunks.

dmitriid:
@warmwaffles That's so true. This was the case when I tried to go in guns a-blazing in https://github.com/dmitriid/sqelect/tree/update-attempt-the-first

It does become weird when you go into the actual ecto adapter. Suddenly you feel like you need 100 things at once (encoding, decoding, oh my). But may be taking things chunk by chunk will work.

warmwaffles:
> It does become weird when you go into the actual ecto adapter. Suddenly you feel like you need 100 things at once (encoding, decoding, oh my). But may be taking things chunk by chunk will work.

Yea that was really confusing. I am trying to figure out parts we can lift out and test appropriately. 

Though as I just worked through trying to make tests pass I was able to decipher what the intentions were.

warmwaffles (approved):

PR #5, opened 2021-02-21
Adds nested transaction support

warmwaffles:
@dmitriid if you can take a look at this and relay your thoughts.


dmitriid (approved):
Looks good! 🚢

dmitriid:
I agree on the weirdness of not handling more than two nested levels

PR #6, opened 2021-02-21
Connection.handle_close/3

dmitriid:
My C-foo is non-existent, so here's what happens (I have `Connection.test/0` as a quick-and-dirty way to test things):

- Run `Connection.test()`, and everything is fine
- Run `Connection.test()` a second time, and:

```
beam.smp(43501,0x7000048e0000) malloc: Non-aligned pointer 0x7fe3598ebc50 being freed (2)
``` 

@warmwaffles


dmitriid:
This PR is now invalid 👍

PR #20, opened 2021-03-01
Add files option to project options (hex package)

Eiji7:
Following https://hex.pm/docs/publish documentation.


warmwaffles:
It is expected that 1 test fails right now. It's an update using a FROM clause.

PR #27, opened 2021-03-01
Add structure_dump/load to support mix ecto.dump/load

### LINKED ISSUE #13: Implement structure dump and load
Currently it is not implemented and needs to be. The implementation should live in `Ecto.Adapters.Exqlite`.

dmitriid:
Fixes #13

I tested this manually with the following:

```elixir

#./config/config.exs

use Mix.Config

config :exqlite,
  ecto_repos: [Exqlite.Repo]

config :exqlite, Exqlite.Repo,
  database: "/<path-to>/test.db"
```

```elixir
defmodule Exqlite.Repo do
  use Ecto.Repo,
    otp_app: :exqlite,
    adapter: Ecto.Adapters.Exqlite
end
```

```elixir
# priv/ecto/migrations/20210301121319_add_user_table.exs

defmodule Exqlite.Repo.Migrations.AddUserTable do
  use Ecto.Migration

  def change do
    create table(:users) do
      add :name, :string
      add :username, :string

      timestamps()
    end

    create unique_index(:users, [:username])
  end
end
```

and then

```
> mix ecto.dump -d ./test_dump.sql
# check the dump. then delete everything from the db

> mix ecto.load -d ./test_dump.sql 
```

Note: we can probably do the same manually by going through the various pragmas, but nah :)

### DIFF
--- a/lib/ecto/adapters/exqlite.ex
+++ b/lib/ecto/adapters/exqlite.ex
@@ -43,8 +43,6 @@ defmodule Ecto.Adapters.Exqlite do
   def structure_dump(default, config) do
     path = config[:dump_path] || Path.join(default, "structure.sql")
 
-    # TODO: dump the database and select the migration versions
-
     with {:ok, contents} <- dump_schema(config),
          {:ok, versions} <- dump_versions(config) do
       File.mkdir_p!(Path.dirname(path))
--- a/lib/ecto/adapters/exqlite.ex
+++ b/lib/ecto/adapters/exqlite.ex
@@ -40,26 +40,28 @@ defmodule Ecto.Adapters.Exqlite do
   def supports_ddl_transaction?(), do: false
 
   @impl Ecto.Adapter.Structure
-  def structure_dump(_default, _config) do
-    # table = config[:migration_source] || "schema_migrations"
-    # path  = config[:dump_path] || Path.join(default, "structure.sql")
-    #
+  def structure_dump(default, config) do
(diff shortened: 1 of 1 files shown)


lib/ecto/adapters/exqlite.ex
```diff
@@ -299,4 +299,36 @@ defmodule Ecto.Adapters.Exqlite do
...
+
+  defp dump_schema(config) do
+    case run_with_cmd("sqlite3", [config[:database], ".schema"]) do
+      {output, 0} -> {:ok, output}
+      {output, _} -> {:error, output}
+    end
+  end
+
+  defp run_with_cmd(cmd, args) do
+    unless System.find_executable(cmd) do
```
warmwaffles, line 327: The sqlite3 release comes with a cli tool. I wonder if we should just package it as well with the binary.

warmwaffles (approved):

lib/ecto/adapters/exqlite.ex
dmitriid: @warmwaffles I'm on the fence about this :) I can se both pros and cons

warmwaffles: Well, if we require them to have an sqlite binary on their machine, I would assume most distros come with their own sqlite3-dev package that we could require them to have. On the other hand, I like this being a single standalone package.

dmitriid: @warmwaffles I just realised that we build sqlite on install anyway. So yeah, why not include the binary

PR #31, opened 2021-03-04
Adds LRU functionality to prepared query cache

### LINKED ISSUE #21: Query cache should be LRU based
Right now the query cache `Exqlite.Queries` currently just grows infinitely. I think the solution we should look at is have this be a configurable value something like `query_cache_size: 100` to set a max value, and setting to `0` would disable query caching.

warmwaffles:
Fixes #21 

This fixes the issue where a memory leak would happen if the connection is held open for a continuous amount of time. Instead we clamp the query cache down to a fixed size specified in an option given to the repo.


dmitriid (approved):

dmitriid:
Ship it :) 👍

PR #36, opened 2021-03-06
Fix stale entry error

warmwaffles:
@dmitriid I could use your help trying to figure this one out. If you check this branch out and run `mix test` you'll see the error.


warmwaffles:
@kevinlang would you mind taking a look at this PR.

PR #39, opened 2021-03-08
Adds QueryCache behaviour

warmwaffles:
@kevinlang I added a behaviour `QueryCache` because I wanted to disable query caching to figure out why this was failing

```elixir
    # this test keeps on hitting busy issues, not sure why
    # one error i saw, tho not sure what test exactly, was
    # passes fine in isolation.
    test "delete_all deletes all products" do
      TestRepo.insert!(%Product{name: "hello"})
      TestRepo.insert!(%Product{name: "hello again"})

      # we have to do this because the tests are not isolated from one another.
      # @kevinlang is working on rectifying that problem
      assert {total, _} = TestRepo.delete_all(Product)
      assert total >= 2
    end
```

And the error comes back as `Database busy` for it.

I found this seed to reliably fail 

```
mix test --seed 383763
```


kevinlang:
Good idea. In original PR I thought it was the QueryCache, then realized that it is not named and is per-connection, so should _theoretically_ be fine. But it does seem like the best bet, even though my attempts of ripping it out did not succeed.

Can confirm that un-commenting out that test and running with that seed does result it that test failing again. I also sometimes get ` test handles streams properly` fail with that rare `cannot open savepoint - SQL statements in progress` issue...

warmwaffles:
@kevinlang I've also thought about just ditching query caching entirely. I don't know how much it's actually helping or hindering by caching.

kevinlang:
AFAIK other standard drivers do not cache statements by default, but I may be mistaken. I don't think `go-sqlite3` does, for example.

Though by "ripping out" above, I meant disabling it temporarily via my own means to test my assumption, but even after doing that (setting cache size to 0 effectively) I did not get any result, kind of like how setting `cache_strategy: :none` here does not seem to help, either.

PR #45, opened 2021-03-09
add ecto and ecto_sql integration tests scaffolding

kevinlang:
Due to our current tests using their own schema, it is difficult to add these tests under `test/` as they require their own schema. 

I see us eventually folding the various ecto-related integration tests into this directory in the future. Keeping them separate for now is fine though, as we try to slowly onboard the large suite of tests in `ecto` and `ecto_sql`.

Just running the schema migration exposed at least one issue, so I can see these being quite helpful as we onboard them.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,3 +21,4 @@ jobs:
           mix deps.get
 
       - run: mix test
+      - run: EXQLITE_INTEGRATION=true mix test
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -21,3 +21,4 @@ jobs:
           mix deps.get
 
       - run: mix test
+      - run: EXQLITE_INTEGRATION=true mix test
```
warmwaffles, line 24: I was going to suggest fanning this one out to be more parallel, but I think this is a good first step.

warmwaffles:
@kevinlang I would be okay with making the schemas uniform with `ecto_sql`.

kevinlang:
In that case I will in a subsequent PR move the existing integration tests for the ecto driver over to this `integration_test/` folder. The `test/` folder will continue to have non-integration tests for the ecto adapter, and integration tests for `exqlite`.

This will make things easier for merging back over to `ecto_sql` when we get to that point, as it reflects the folder structure they have.

Sounds good?

warmwaffles:
Yea, I'm working on multi step right now.

Issue #42, opened 2021-03-08
Database busy error


warmwaffles:
@kevinlang I am going to track the database busy issue here.

Previous discussions:

* https://github.com/warmwaffles/exqlite/pull/38
* https://github.com/warmwaffles/exqlite/pull/39#issuecomment-792447244

warmwaffles:
I need to dig into `Ecto.Adapters.SQL.Sandbox` more to ascertain if it can work with sqlite's restriction of 1 writer at a time. We are definitely hitting some issues where another test has opened a transaction but has not committed it yet and we are trying to read back those changes.

Yay for concurrency issues.

kevinlang:
It is worth noting that the tests are running sequentially, and that the sandbox makes it so one checks out only a single connection at a time. So the issue we are hitting in our tests is more sequential than a real world scenario.

I think you are right that there is some issue where a sandboxed connection is not closing out correctly, affecting the other tests.

One other approach I thought of was to run tests against an in-memory database. At "checkout" time we would clone the on-disk database into a `:memory:` one. There we don't have to worry about cleanup or anything, assuming check-in drops the "connection" and thus the in-memory database.

However, I kind of like that the current way the sandbox does it, with transaction, raises these trickier transaction issues in our test suite, which perhaps reflect the real world a bit more than the memory approach.

warmwaffles:
> However, I kind of like that the current way the sandbox does it, with transaction, raises these trickier transaction issues in our test suite, which perhaps reflect the real world a bit more than the memory approach.

This is precisely why I strayed away from using `:memory:` for integration tests.

Perhaps, instead of just returning `{:error, "Database busy"}` we can retry the statement again after a random wait?

https://github.com/warmwaffles/exqlite/blob/98dade3a5996d6930e3928eba6501a60e112db1e/lib/exqlite/connection.ex#L291-L292

warmwaffles:
The current test error originates from this line

https://github.com/warmwaffles/exqlite/blob/98dade3a5996d6930e3928eba6501a60e112db1e/lib/exqlite/sqlite3.ex#L105

I changed it to 

```elixir
:busy -> fetch_all(conn, statement)
```

And it just hung there

```
23:06:33.422 [error] Exqlite.Connection (#PID<0.434.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.703.0> timed out because it queued and checked out the connection for longer than 15000ms
```

warmwaffles:
[...]

PR #48, opened 2021-03-10
fix issue where we were using same statement ref across db connections

kevinlang:
addresses #42


kevinlang:
@warmwaffles this fixes the issue in the sense that the repo case no longer fails. Not sure _how_ we are getting a `ref` at all though, trying to look through the `db_connection` machinery to figure out...

warmwaffles:
But this looks perfectly reasonable to me. I can't get tests to fail on my machine either.

PR #50, opened 2021-03-10
Adds busy_timeout configuration fixes #49

### LINKED ISSUE #49: Make busy_timeout configurable
Useful to configure in itself, but also for our tests that rely on testing it. It makes those tests hang for ~2seconds per the `2000` value we have hardcoded.

warmwaffles:


kevinlang (approved):

warmwaffles:
I thought about exposing another NIF function for this, but it can be simply done with a pragma.

PR #52, opened 2021-03-10
onboard ecto_sql sql.exs integration test, various fixes

kevinlang:
Onboarded 17 new tests! :)

### DIFF
--- a/integration_test/exqlite/test_helper.exs
+++ b/integration_test/exqlite/test_helper.exs
@@ -1,6 +1,5 @@
 Logger.configure(level: :info)
 
-Application.put_env(:ecto, :json_library, Jason)
 Application.put_env(:ecto, :primary_key_type, :id)
 Application.put_env(:ecto, :async_integration_tests, false)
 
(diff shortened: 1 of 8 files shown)


lib/exqlite/connection.ex
```diff
@@ -439,6 +439,12 @@ defmodule Exqlite.Connection do
 
   defp maybe_changes(_, _), do: nil
 
+  # TODO: should we change the NIF to just return the string
+  defp normalize_columns(columns) do
```
kevinlang, line 443: @warmwaffles I'm curious your thoughts here. I think we need the extra info for help in decoding the results, but can't find where that lives, and making this change didn't result in any tests failing.

We may need to do this normalization only within the Ecto layer, instead.

warmwaffles: Yea we can make that change in the nif. It's fairly easy.

warmwaffles:
Let me add that to this PR real quick

warmwaffles:
@kevinlang made a quick commit where `columns/2` just returns the column names in a list.

integration_test/exqlite/test_helper.exs
```diff
@@ -41,9 +56,15 @@ _ = Ecto.Adapters.Exqlite.storage_down(TestRepo.config())
...
 
 :ok = Ecto.Migrator.up(TestRepo, 0, Ecto.Integration.Migration, log: false)
 Ecto.Adapters.SQL.Sandbox.mode(TestRepo, :manual)
 Process.flag(:trap_exit, true)
 
-ExUnit.start()
+ExUnit.start(
+  exclude: [
+    # SQLite does not have an array type
+    :array_type,
```
warmwaffles, line 68: It doesn't have an array type, but I specified it to get serialized as a json array. Not sure if this is just for the postgres specific array operation stuff.

lib/exqlite/sqlite3.ex
```diff
@@ -113,5 +113,6 @@ defmodule Exqlite.Sqlite3 do
   defp convert(%DateTime{} = val), do: DateTime.to_iso8601(val)
   defp convert(%Time{} = val), do: Time.to_iso8601(val)
   defp convert(%NaiveDateTime{} = val), do: NaiveDateTime.to_iso8601(val)
+  defp convert(nil), do: "NULL"
```
warmwaffles, line 116: Oof, how did I miss this.

warmwaffles (approved):

integration_test/exqlite/test_helper.exs
warmwaffles: Yea there is no way we can support that. Unfortunate.

PR #56, opened 2021-03-10
support Decimal type

petermm:
using :decimal gave this error:
```

[debug] QUERY ERROR db=12.5ms queue=0.1ms idle=1882.2ms
INSERT INTO measurements (sensor_name,type,value,inserted_at,updated_at) VALUES (?,?,?,?,?) ["room", "temp", #Decimal<13.3>, "2021-03-10T12:10:44", "2021-03-10T12:10:44"]
[error] GenServer #PID<0.511.0> terminating
** (DBConnection.ConnectionError) client #PID<0.696.0> stopped: ** (WithClauseError) no with clause matching: {:error, %Exqlite.Error{message: {:wrong_type, #Decimal<13.3>}}, %Exqlite.Connection{db: #Reference<0.2463941217.3987865620.167132>, path: "dev.sqlite3", status: :busy, transaction_status: :idle}}
```

copied from https://github.com/elixir-sqlite/sqlitex/blob/master/lib/sqlitex/statement.ex#L274

PR only handles Decimal - was not sure of other types..

no tests - assume larger test suite is being pulled in, else let me know and I'll add tests..

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -487,6 +487,19 @@ defmodule Exqlite.Connection do
     #      https://www.sqlite.org/c3ref/bind_blob.html E.g. we can accept a map of params
     #      that binds values to named params. We can look up their indices via
     #      https://www.sqlite.org/c3ref/bind_parameter_index.html
+
+    params =
+      Enum.map(params, fn
+        # nil -> :undefined
+        # true -> 1
+        # false -> 0
+        # date = {_yr, _mo, _da} -> date_to_string(date)
+        # time = {_hr, _mi, _se, _usecs} -> time_to_string(time)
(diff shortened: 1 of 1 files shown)


warmwaffles:
@petermm can you add a small test explicitly for this? The larger test suite is an in progress sort of thing.

warmwaffles:
@petermm I think the thing to do actually here is to just use `Decimal.to_string(value, :normal)` because decimals are stored in sqlite as `NUMERIC` but have a resulting affinity of `TEXT`. So it's probably simpler to just do that above. I can do this in a separate PR.

warmwaffles:
I see. Okay, I'll start work on that now. I think @petermm was on to something.

warmwaffles:
@petermm thank you for bringing this to my attention. I went ahead and implemented it on the connection side for the ecto adapter in #59. I'm going to close this since that one handles different cases better.

PR #59, opened 2021-03-10
Handle decimals

warmwaffles:
@kevinlang this is a different approach from #56 

If you uncomment this

```
# Code.require_file "#{ecto}/integration_test/cases/type.exs", __DIR__
```

And run 

```
EXQLITE_INTEGRATION=true mix test --only decimal_type --seed 0
```

You'll see that it at least returns numerically correct values. The biggest issue right now is that this happens

```
  3) test on coalesce with mixed types (Ecto.Integration.TypeTest)
     deps/ecto/integration_test/cases/type.exs:444
     match (=) failed
     The following variables were pinned:
       decimal = #Decimal<1.0>
     code:  assert [^decimal] = TestRepo.all(from(p in Post, select: coalesce(p.cost, 0)))
     left:  [^decimal]
     right: [1]
     stacktrace:
       deps/ecto/integration_test/cases/type.exs:447: (test)
```

It's being returned as an integer and another one is being returned as a normalized `Decimal` value.

### DIFF
--- a/lib/ecto/adapters/exqlite.ex
+++ b/lib/ecto/adapters/exqlite.ex
@@ -5,6 +5,8 @@ defmodule Ecto.Adapters.Exqlite do
   @behaviour Ecto.Adapter.Storage
   @behaviour Ecto.Adapter.Structure
 
+  alias Ecto.Adapters.Exqlite.Codec
+
   @impl Ecto.Adapter.Storage
   def storage_down(options) do
     db_path = Keyword.fetch!(options, :database)
@@ -101,7 +103,7 @@ defmodule Ecto.Adapters.Exqlite do
 
   @impl Ecto.Adapter
   def loaders(:boolean, type) do
-    [&bool_decode/1, type]
+    [&Codec.bool_decode/1, type]
   end
 
   @impl Ecto.Adapter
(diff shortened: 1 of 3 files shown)


warmwaffles:
This is something with how sqlite handles decimals. It just returns the sum as a integer if there is nothing past the decimal

```
$ sqlite3 :memory:
SQLite version 3.34.1 2021-01-20 14:10:07
Enter ".help" for usage hints.
sqlite> create table foo (id integer primary key, cost decimal(10, 2));
sqlite> insert into foo (id,cost) values (1, '1.00');
sqlite> select sum(foo.cost) + 0 from foo;
1
```

warmwaffles:
@kevinlang I got this commit into ecto proper https://github.com/elixir-ecto/ecto/commit/63068b50cf1762a2a0b22a16e1aa551e1dac0d5c

We just need to wait for it to get released into an official package. But this PR with the changes to the deps/ dir pass now.

kevinlang (approved):

kevinlang:
Nice! We will probably end up "vendor"ing that `type.exs` file into the `integration_test/` directory similar to how we did with the `sandbox.exs` one.

warmwaffles:
Yea I intended to have extra and different tests for exqlite. Just because the database functions differently than others.

PR #61, opened 2021-03-10
onboard logging.exs integration test

kevinlang:


warmwaffles:
Heh always okay with more tests being added.

PR #70, opened 2021-03-12
add support for ANY via JSON_EACH

kevinlang:


warmwaffles:
What an interestingly clever solution to substitute for `ANY`

warmwaffles:
@kevinlang can we get a test added to `ecto/adapters/connection_test.exs` for it? I remember either out right deleting the original or just asserting that a raise happened.

kevinlang:
@warmwaffles added a test :+1:

PR #73, opened 2021-03-12
 onboard interval.exs integration tests 

dozoyo:
Since SQLite does not support microsecond precision, I've forked the interval tests for now and duplicated the ones that use microseconds to instead use milliseconds when validating different datetime types.

The reason I changed the "DATETIME" type to "TEXT_DATETIME" is that "DATETIME" would truncate the date and cast it incorrectly:
```
sqlite> SELECT strftime('%Y-%m-%d %H:%M:%f',CAST("2014-01-01 02:00:00.001501" AS DATETIME));
-470-05-30 12:00:00.000
```
With TEXT_DATETIME:
```
sqlite> SELECT strftime('%Y-%m-%d %H:%M:%f',CAST("2014-01-01 02:00:00.001501" AS TEXT_DATETIME));
2014-01-01 02:00:00.002
```

The Z addition is what's needed to load in the other datetime types. Otherwise Ecto fails to recognize and cast it.
And the parentheses are needed for sqlite to properly parse additions, otherwise it just grabs the first number.

The last thing is that I found a previously commented out test now works. I believe it was due to the resolved database_busy errors.

### DIFF
--- a/integration_test/exqlite/all_test.exs
+++ b/integration_test/exqlite/all_test.exs
@@ -17,6 +17,10 @@ Code.require_file "#{ecto_sql}/integration_test/sql/logging.exs", __DIR__
 # necessary to fork into the repo as it relies on :ecto_sql app, which
 # we modified to be :exqlite app until we merge into ecto_sql
 Code.require_file "./ecto_sql/sandbox.exs", __DIR__
+# Since sqlite does not have microsecond precision we forked these tests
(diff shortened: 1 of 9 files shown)


warmwaffles:
@dozoyo Also may want to push that upstream as well! I was looking at this yesterday and saw that it could use some extra tagging there.

Issue #80, opened 2021-03-12
Improve on_conflict: nothing and un-fork repo.exs tests


kevinlang:
Right now we have a caveat about us returning an invalid PK id when hitting a conflict during `ON CONFLICT DO NOTHING`.

This is due to limitations around the last insert id, as mentioned in the README.

This also caused us to have to fork the `repo.exs` cases so that we can exclude one of the tests around this particular scenario.

Once #76 is supported, and we use that as the primary mechanism for getting the last insert ids, we should be able to eliminate this caveat and un-fork our repo.exs test.

I was able to confirm that the `RETURNING *` clause returns nothing in the `ON CONFLICT DO NOTHING` case, which is what we want.
```
create table foo(bar);
create unique index u on foo(bar);
insert into foo values(1) returning *;
=> 1
insert into foo values(1) on conflict do nothing returning *;
(nothing returned)
```

PR #81, opened 2021-03-12
Adds RETURNING support

### LINKED ISSUE #76: Add support for RETURNING clause
Now that we have 3.35.0 we can support this.

see #75

warmwaffles:
closes #76

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,10 +16,6 @@ An SQLite3 library with an Ecto adapter implementation.
   second column somewhere storing the timezone name and shifting it when you
   get it from the database. This is more reliable than storing the offset as
   `+03:00` as it does not respect daylight savings time.
-* `on_conflict: :ignore` may return an invalid / non-sensical primary key
-  if a conflict was indeed hit. This is due to limitations around not being able
-   to detect a conflict happened and thus using the last row insert id 
-   which could be from any table.
 
 
 ## Installation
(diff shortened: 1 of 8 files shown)


warmwaffles:
@kevinlang this just generates the SQL necessary to support returning.

Is there an integration in ecto_sql that we can pull in for this?

warmwaffles:
Found it, it's the `returning` tag being excluded.

kevinlang:
`function nil.encode/1 is undefined` is due to the `:json_library` thing. 

I think what we will need to do, to get rid of the warning, is adopt the pattern in other adapters which do:
```
 Application.get_env(:postgrex, :json_library, Jason)
 ```
 Everytime they access the json library to avoid our needing to set it in test setup

warmwaffles:
Yea I feel ya there. Stupid warning. I'll revert that.

warmwaffles:
@kevinlang I have 1 test failing in the integration test suite. This also flips back over to the provided repo integration test

warmwaffles:
I rebased it off of main. I have to head out for the evening and if you find a fix go ahead and push it into here and merge away.

kevinlang:
Nice work! For #80 , I'm not sure if there is a clean solution. I was looking through and assumed that Postgrex and others roguely tacked on `RETURNING` clauses under the hood to get insert ids, but that doesn't seem to be the case. I can't find any solution actually to the `:on_conflict_ignore` thing, am I missing something?

kevinlang:
The fix was just a matter of removing a bunch of our custom `insert` "override" code. 

Now the ecto layer no longer has any more need for the last insert id. We may want to add some way to avoid making the `last_insert_id` call when using the lower-level driver with ecto, to eek out some extra perf.

PR #92, opened 2021-03-13
fix json_encode

### LINKED ISSUE #60: json_extract_path integration tests fail
If you uncomment `Code.require_file "#{ecto}/integration_test/cases/type.exs", __DIR__` and run 

```
EXQLITE_INTEGRATION=true mix test --only json_extract_path --seed 0
```

You'll see the errors.

kevinlang:
A couple of issues:

1. We were not encoding JSON for the `:map` type.
2. We were wrapping the various json keys used in lookup in double quotes, which is not necessary in SQLite
3. We were opting-in to a binary key that `Jason` would fail to encode

(2) manifests itself mainly here:
```
select json_extract('{ "a": 1, "\"b\"": 2 }', '$.a');
=> 1

select json_extract('{ "a": 1, "\"b\"": 2 }', '$."a"');
=> 1

select json_extract('{ "a": 1, "\"b\"": 2 }', '$.\"b\"');
=> 2

select json_extract('{ "a": 1, "\"b\"": 2 }', '$."\"b\""');
(no result)
```

That last row was causing us issues, and the solution was to simply not wrap with double quotes to avoid it. It may be a bug in SQLite, I think. I started a thread about it [here](https://sqlite.org/forum/forumpost/1f5263a98f).

Fixes #60

### DIFF
--- a/lib/ecto/adapters/exqlite.ex
+++ b/lib/ecto/adapters/exqlite.ex
@@ -80,11 +80,6 @@ defmodule Ecto.Adapters.Exqlite do
   end
 
   @impl Ecto.Adapter
-  def loaders(:binary_id, type) do
-    [Ecto.UUID, type]
-  end
-
-  @impl Ecto.Adapter
   def loaders(:naive_datetime_usec, type) do
     [&Codec.datetime_decode/1, type]
   end
@@ -167,11 +162,6 @@ defmodule Ecto.Adapters.Exqlite do
   end
 
   @impl Ecto.Adapter
-  def dumpers(:binary_id, type) do
-    [type, Ecto.UUID]
-  end
-
-  @impl Ecto.Adapter
   def dumpers(:boolean, type) do
     [type, &Codec.bool_encode/1]
   end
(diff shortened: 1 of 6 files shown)


lib/ecto/adapters/exqlite.ex
```diff
@@ -79,11 +79,6 @@ defmodule Ecto.Adapters.Exqlite do
     [&Codec.bool_decode/1, type]
   end
 
-  @impl Ecto.Adapter
-  def loaders(:binary_id, type) do
```
warmwaffles, line 83: Is this no longer necessary to specify? or was this causing an unrelated bug?

kevinlang: Without this removed, I was hitting a `Jason` encoding error related to the UUID being provided as a (not sure of terminology) charlist. E.g.,
```
<<41, 220, 200, 189, 24, 33, 72, 249, 188, 18, 214, 5, 143, 148, 203, 28>>
```

Instead of, say "407ef7a9-0fd9-4abf-a5bc-22b7c12f0884"

warmwaffles: Oh I see. Let's remove it then and make a follow on issue to ensure that UUID is functioning as advertised and what not.

warmwaffles (approved):

Issue #105, opened 2021-03-15
Onboard "embed with custom type" test by fixing issue related to Ecto select projection


kevinlang:
This is the last test from `type.exs` that is marked with our `:onboarding` tag.

See notes in #101. I think that Ecto may have an issue where if the query has a `select` for a specific column, it results in the respective loader not getting called, which is why that test is failing. This issue would manifest itself pretty frequently in day-to-day usage if my theory is correct - everytime a specific column is `select`ed that would benefit from a loader pass, which in SQLite is most of them!

Edit: also if my theory is correct, the fix for this is upstream

kevinlang:
PR #106 has an example of what I mean. Looks like we don't do sufficient loader invocation when building the query via the table string name (e.g., "users") instead of the Ecto schema (e.g., `User`).

kevinlang:
Closed in 10706d1

PR #106, opened 2021-03-15
add simple repro case for #105

kevinlang:
```
  1) test insert insert user w/ meta (Ecto.Integration.CrudTest)
     test/ecto/integration/crud_test.exs:13
     Assertion with == failed
     code:  assert [user1.meta] == TestRepo.all(from(u in "users", select: u.meta))
     left:  [%{a: "b"}]
     right: ["{\"a\":\"b\"}"]
     stacktrace:
       test/ecto/integration/crud_test.exs:15: (test)
```

### DIFF
--- a/test/ecto/integration/crud_test.exs
+++ b/test/ecto/integration/crud_test.exs
@@ -10,6 +10,11 @@ defmodule Ecto.Integration.CrudTest do
   import Ecto.Query
 
   describe "insert" do
+    test "insert user w/ meta" do
+      {:ok, user1} = TestRepo.insert(%User{name: "John", meta: %{a: "b"}}, [])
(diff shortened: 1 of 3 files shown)


warmwaffles:
Schemaless queries are supposed to fail like that right? Ecto doesn't know that `users` is `User`, thus you are supposed to pass the module definition with the schema along.

kevinlang:
Good point @warmwaffles ! I found this piece of doc that helped clear things up for me a bit more too: 

https://hexdocs.pm/ecto/schemaless-queries.html

warmwaffles:
Yea I remember seeing some tests in `connection_test` that explicitly checks for certain things that are not possible with a schemaless query.

PR #112, opened 2021-03-17
onboard transaction.exs suite, various transaction fixes

### LINKED ISSUE #58: Sandbox does not recover properly from errors, leading to busy issues
If you go to `integration_test/exqlite/all_test.exs` and comment out the line for:

```
Code.require_file "#{ecto}/integration_test/cases/interval.exs", __DIR__
```

and run the test suite, the first test fails due to some reason, and then the rest of the tests fail with "Database busy". Which means the failed test is not rolling back its transaction correctly...

I believe `db_connection` has a `:disconnect` return status that I noticed some of the other adapters take use of when hitting errors. I think we need to use that for some of the errors we cannot recover from, so that the entire connection is closed and the database does not stay in an unreachable state until the entire application is restarted!

To be clear, this may not be scoped only to the sandbox. This probably represents an error that could occur in normal "production" application use.

### LINKED ISSUE #93: Onboard transaction.exs test suite
This is the most reliable reproduction of #58. My bet is that once we have all the tests passing here, #58 will be fine, too.

Labeling as bug since these failing tests indicate an underlying bug that will be fixed at the same time the tests are onboarded.

kevinlang:
The main test that was failing was this one, which was subsequently blocking the other tests was:

```
    @tag :transaction_checkout_raises
    test "checkout raises on transaction attempt" do
      assert_raise DBConnection.ConnectionError, ~r"connection was checked out with status", fn ->
        PoolRepo.checkout(fn -> PoolRepo.query!("BEGIN") end)
      end
    end
```

We assume that all transaction-beginning statements go through our `handle_transaction` call which is not the case. Here we don't detect we begin a transaction and thus don't know we are in an active transaction at "checkin" time, hence why the error is not thrown.

The fix is to add a new `transaction_status` NIF call that uses http://www.sqlite.org/c3ref/get_autocommit.html to return the current transaction status, and wiring that up in the correct places.

The second issue was that `:disconnect` was not properly rolling back a transaction. This is because `sqlite3_close_v2` is not guaranteed to succeed, and indeed was going into a "zombie" state, as described at https://www.sqlite.org/c3ref/close.html . The fix for this was to make sure we atleast roll back any active transaction before doing this close logic, and then relying on the GC to clean the open prepared statements.

This closes #93 and likely also closes #58 .

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -572,7 +572,7 @@ exqlite_transaction_status(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     int autocommit = sqlite3_get_autocommit(conn->db);
     return make_ok_tuple(
         env,
-        autocommit == 0 ? enif_make_atom(env, "transaction") : enif_make_atom(env, "idle"));
+        autocommit == 0 ? make_atom(env, "transaction") : make_atom(env, "idle"));
 }
 
 static void
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -145,6 +145,7 @@ exqlite_close(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     assert(env);
 
     connection_t* conn = NULL;
+    int rc = SQLITE_OK;
 
     if (argc != 1) {
         return enif_make_badarg(env);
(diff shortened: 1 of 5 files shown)


warmwaffles (approved):
Interesting find. Thank you for leaving that note there.

PR #125, opened 2021-03-22
upgrade CI, format

kevinlang:
This is adopted from the pattern Dashbit uses. E.g.,

https://github.com/dashbitco/mox/blob/master/.github/workflows/ci.yml

CI should run a bit faster now, and this setup will also allow us to leverage the improvements of the underlying action that is supported by the Erlang Foundation.

Original intent was to do this for Windows support, but it looks like it does not support Windows yet :(

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   test:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -1,23 +1,47 @@
...
-jobs:
-  test-elixir:
-    runs-on: ubuntu-latest
-    container: elixir:${{ matrix.elixir_version }}-slim
 
+on:
+  pull_request:
+  push:
+    branches:
+      - master
```
cw789, line 7: Default branch is `main`.

cw789 (commented):
Probably overseen from the reference.

.github/workflows/ci.yml
warmwaffles: good catch

PR #126, opened 2021-03-23
Improve guide for Windows users

cw789:
Extend section for Visual Studio Code users.

### DIFF
--- a/guides/windows.md
+++ b/guides/windows.md
@@ -1,44 +1,81 @@
-# Windows Users
+# Windows users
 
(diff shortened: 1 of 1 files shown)


kevinlang:
Thanks!

cw789:
> Thanks!

No need to say. It's anyway only a note for me (as long as I don't escape Windows into virtual machines). 😁
Anyway, this is probably now the most up-to-date guide for any library using NIFs, to be used on Windows.

PS: Found one typo, will fix that later.

Issue #127, opened 2021-03-23
When connection times out a SIGSEGV is encountered


warmwaffles:
For a large table of 66,000 records

```elixir
Company
|> select([c], c)
|> Repo.all()
```

Encounter 

```
14:25:34.215 [error] Exqlite.Connection (#PID<0.446.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.536.0> timed out because it queued and checked out the connection for longer than 15000ms

#PID<0.536.0> was at location:

    (exqlite 0.5.1) lib/exqlite/sqlite3.ex:107: Exqlite.Sqlite3.fetch_all/3
    (exqlite 0.5.1) lib/exqlite/connection.ex:486: Exqlite.Connection.get_rows/2
    (exqlite 0.5.1) lib/exqlite/connection.ex:439: Exqlite.Connection.execute/4
    (db_connection 2.3.1) lib/db_connection/holder.ex:316: DBConnection.Holder.holder_apply/4
    (db_connection 2.3.1) lib/db_connection.ex:1272: DBConnection.run_execute/5
    (db_connection 2.3.1) lib/db_connection.ex:1359: DBConnection.run/6
    (db_connection 2.3.1) lib/db_connection.ex:557: DBConnection.parsed_prepare_execute/5
    (db_connection 2.3.1) lib/db_connection.ex:550: DBConnection.prepare_execute/4
    (ecto_sqlite3 0.5.3) lib/ecto/adapters/sqlite3/connection.ex:62: Ecto.Adapters.SQLite3.Connection.prepare_execute/5
    (ecto_sql 3.5.4) lib/ecto/adapters/sql.ex:692: Ecto.Adapters.SQL.execute!/4
    (ecto_sql 3.5.4) lib/ecto/adapters/sql.ex:684: Ecto.Adapters.SQL.execute/5
    (ecto 3.5.8) lib/ecto/repo/queryable.ex:229: Ecto.Repo.Queryable.execute/4
    (ecto 3.5.8) lib/ecto/repo/queryable.ex:17: Ecto.Repo.Queryable.all/3
    (stdlib 3.14) erl_eval.erl:680: :erl_eval.do_apply/6
    (elixir 1.11.3) src/elixir.erl:280: :elixir.recur_eval/3
    (elixir 1.11.3) src/elixir.erl:265: :elixir.eval_forms/3
[...]

PR #128, opened 2021-03-23
Adds multi_step/3

warmwaffles:
Related to the error I encountered with #127.

This does not fix the issue where a timeout causes a `sigsegv`. This just kicks the can further down the road.

This should also cause ecto_sqlite3 to bump this as a dependency to pickup the next change.

### DIFF
--- a/lib/exqlite/sqlite3.ex
+++ b/lib/exqlite/sqlite3.ex
@@ -112,12 +112,12 @@ defmodule Exqlite.Sqlite3 do
     fetch_all(conn, statement, [])
   end
 
-  defp fetch_all(conn, statement, result) do
+  defp fetch_all(conn, statement, accum) do
     case multi_step(conn, statement) do
       :busy -> {:error, "Database busy"}
       {:error, reason} -> {:error, reason}
-      {:done, rows} -> {:ok, result ++ Enum.reverse(rows)}
-      {:rows, rows} -> fetch_all(conn, statement, result ++ Enum.reverse(rows))
+      {:done, rows} -> {:ok, accum ++ Enum.reverse(rows)}
+      {:rows, rows} -> fetch_all(conn, statement, accum ++ Enum.reverse(rows))
     end
   end
 
--- a/lib/exqlite/sqlite3.ex
+++ b/lib/exqlite/sqlite3.ex
@@ -19,6 +19,7 @@ defmodule Exqlite.Sqlite3 do
   @type db() :: reference()
   @type statement() :: reference()
   @type reason() :: atom() | String.t()
+  @type row() :: []
 
   @doc """
   Opens a new sqlite database at the Path provided.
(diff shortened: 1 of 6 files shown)


warmwaffles:
Working through the results being accumulated out of order

lib/exqlite/sqlite3.ex
```diff
@@ -78,6 +78,27 @@ defmodule Exqlite.Sqlite3 do
   @spec step(db(), statement()) :: :done | :busy | {:row, []}
   def step(conn, statement), do: Sqlite3NIF.step(conn, statement)
 
+  @spec multi_step(db(), statement()) :: :busy | {:rows, [[]]} | {:done, [[]]}
+  def multi_step(conn, statement), do: multi_step(conn, statement, 50)
```
kevinlang, line 82: can we make the chunk size configurable?

lib/exqlite/sqlite3.ex
```diff
@@ -78,6 +78,27 @@ defmodule Exqlite.Sqlite3 do
...
+  def multi_step(conn, statement, chunk_size) do
+    case Sqlite3NIF.multi_step(conn, statement, chunk_size) do
+      :busy ->
+        {:error, "Database busy"}
+
+      {:error, reason} ->
+        {:error, reason}
+
+      {:rows, rows} ->
+        {:rows, Enum.reverse(rows)}
```
kevinlang, line 95: can you leave a comment as to why we need the call to `Enum.reverse` here? I'm not entirely sure why it is needed (though it looks like it is indeed, by virtue of the test passing)

lib/exqlite/sqlite3.ex
kevinlang: I can see a higher chunk size being nice in environments where one is not concerned about startving out the dirty NIF scheduler (e.g., when there is a large core count)

lib/exqlite/sqlite3.ex
kevinlang: guessing its a NIF thing

kevinlang (approved):
two comments about adding a note about the reverse and making it configurable so e.g., Ecto adapter users can tune if needed (or the adapter itself can work on finding a nice default value)

lib/exqlite/sqlite3.ex
warmwaffles: Because of the nature of the linked lists that are in erlang, it's easier to just keep prepending in the nif. The other solution would be to allocate the chunk size in the nif and set the values there. Then when it is done, we call `enif_make_list_from_array`.

Pros and cons to both ways. Allocating the array and then calling `enif_make_list_from_array` requires us to manually free the underlying array before returning. Not a huge deal.

lib/exqlite/sqlite3.ex
warmwaffles: I think we could do that. Although for ecto_sqlite3, I'd like that to be configurable per connection and not just for all connections.

warmwaffles:
@kevinlang at minimum I added the `default_chunk_size`. I'll do some fiddling for a follow on PR.

lib/exqlite/sqlite3.ex
kevinlang: For per-connection, I'm not sure how that would look. The only way I can think of making it configurable like that is by hooking into `:configure` and having it per index, but not sure how that would be useful. What were you picturing?

warmwaffles: Sorry I meant to say per repo.

```elixir
  use Ecto.Repo,
    otp_app: :myapp,
    adapter: Ecto.Adapters.SQLite3,
    chunk_size: 1000
```

stuff like that.

kevinlang: got it, that makes sense. that's what I was thinking, too :+1:

Issue #138, opened 2021-04-05
Database is locked investigation


kevinlang:
I still see this now and again in tests, and generally only when the Repo is first set up and connections are getting initialized.

I added a line to `Sqlite3.execute` to log the SQL when it hits an error from the NIF, and was able to find that the error is only hit when the `PRAGMA JOURNAL_MODE=WAL` statement is executed.

My initial thought around this issue was it being an issue around _opening_ the literal database connection, but now I see it is an issue with _configuring_ the database connection after the actual db handle is open.

I was able to replicate this with my local `sqlite3` with the following command:

```
sqlite3 "wow10.db" "PRAGMA JOURNAL_MODE=WAL" 
& sqlite3 "wow10.db" "PRAGMA JOURNAL_MODE=WAL" 
& sqlite3 "wow10.db" "PRAGMA JOURNAL_MODE=WAL" 
( and so on )
```

Intermittently getting the same "database is locked" error.

Not sure what the long term solution here is. Looks like concurrently setting the journal mode can result in a database is locked issue. 

My first thought was to make the journal mode set at compile time, since it is the "default" for `exqlite` anyway, but that doesn't seem to be a compile time option.

If I cannot think of anything I'll probably create a thread in the SQLite3 forum to see if they have a proposed solution.

warmwaffles:
So I was actually thinking about this the other day.

When `WAL` is enabled, it stays enabled even after the sqlite handle is closed. So what we _could_ do is work in a check and set method instead of just blanket setting everything.

kevinlang:
With that PR and also with https://github.com/elixir-sqlite/ecto_sqlite3/pull/25 I no longer see this issue. I think we found the root cause and have the fixes in place, closing.

PR #139, opened 2021-04-05
Adds .maybe_set_pragma/3

warmwaffles:


kevinlang:
I don't think all PRAGMAs are persistent, and some are expected to be required during setup at each connection:

* http://sqlite.1065341.n5.nabble.com/Which-pragmas-are-persistent-td95287.html
* https://stackoverflow.com/questions/55158295/are-sqlite-pragma-settings-permanent-or-do-they-disapear-after-db-close
* https://www.sqlite.org/fileformat2.html#the_database_header

I think we only need to do this for PRAGMAs that modify the database header. By cross-referencing the fileformat docs with the [pragma docs](https://www.sqlite.org/pragma.html), I think the only really relevant one is the `journal_mode` one, as even the `cache_size` one, for example, is not persistent (only deprecated `default_cache_size` actually modifies the corresponding header value).

Can we change this to just do this for the `journal_mode` pragma, for now?

warmwaffles:
Yeah I'm okay with that. I'll make the change soon.

warmwaffles:
@kevinlang I reverted the other `maybe_set_pragma`s and only apply it for `journal_mode` and `cache_size`. This is probably going to help with #138 some.

I wish the db connection pool on the ecto side would have the ability to specify how many of the pool can be writers with the number of read only. Probably difficult to implement.

kevinlang:
Looks good, but I don't think we need it for `cache_size` - it is set per-connection.

See https://www.sqlite.org/pragma.html, which mentions only `default_cache_size` (deprecated) is persisted/shared across all connections.

kevinlang (approved):

kevinlang:
I think we may still hit the "database is locked" issue with these changes the first time the database is ever accessed, since we have no mutex and thus could have a race condition. But it should fix it for all subsequent runs.

We can maybe fix the first-run scenario by having the ecto adapter `storage_up` which calls `Exqlite.Sqlite3.open` to perhaps instead call `Exqlite.Connection.connect`, to do this setup at that time which I _think_ only uses one connection. Will note for later

warmwaffles:
> We can maybe fix the first-run scenario by having the ecto adapter storage_up which calls Exqlite.Sqlite3.open to perhaps instead call Exqlite.Connection.connect, to do this setup at that time which I think only uses one connection. Will note for later

That's actually a pretty good idea

PR #146, opened 2021-04-29
fix issue with update returning nil rows for empty returning result

### LINKED ISSUE #147: `update_all` with select raises error when no rows are updated
The ecto spec for `Ecto.Repo.Assoc.query` expects a list of rows but currently the `ecto_sqlite3` adapter returns `nil` when there were no rows updated from the query.

https://github.com/elixir-ecto/ecto/blob/master/lib/ecto/repo/assoc.ex#L10

This results in the following error:

```bash
** (Protocol.UndefinedError) protocol Enumerable not implemented for nil of type Atom. This protocol is implemented for the following type(s): Ecto.Adapters.SQL.Stream, Exqlite.Stream, DBConnection.Stream, DBConnection.PrepareStream, Timex.Interval, List, Date.Range, Stream, Range, Map, HashDict, GenEvent.Stream, MapSet, HashSet, IO.Stream, Function, File.Stream
```

P.S. Thank you so much for making this adapter! I looked into creating it a year ago but didn't know where to start :1st_place_medal:

kevinlang:
closes #147

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -471,8 +471,11 @@ defmodule Exqlite.Connection do
 
   defp maybe_changes(_, _), do: nil
 
-  defp maybe_rows([]), do: nil
-  defp maybe_rows(rows), do: rows
+  # when we have an empty list of columns, that signifies that
+  # there was no possible return tuple (e.g., update statement without RETURNING)
+  # and in that case, we return nil to signify no possible result.
+  defp maybe_rows([], []), do: nil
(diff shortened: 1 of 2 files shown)


warmwaffles:
@kevinlang when I get time I'll jump in on this later today. I remember running into something like this for `insert_all`.

kevinlang:
@warmwaffles I updated the PR with a fix for this, turns out no nif changes were needed.

warmwaffles (approved):

PR #148, opened 2021-05-03
serialize and deserialize functions

mbuhot:
This PR adds bindings for the [sqlite3_serialize](https://sqlite.org/c3ref/serialize.html) and [sqlite3_deserialize](https://sqlite.org/c3ref/deserialize.html) functions.

The intended use case is to implement an Ecto sandbox that will deserialize the test database into a temporary memory database on checkout, as suggested by @kevinlang on the forum (https://elixirforum.com/t/ecto-sqlite3-an-ecto3-sqlite3-adapter/38391/26)


c_src/sqlite3_nif.c
```diff
@@ -632,6 +632,73 @@ exqlite_transaction_status(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    if (!enif_get_resource(env, argv[0], connection_type, (void**)&conn)) {
+        return make_error_tuple(env, "invalid_connection");
+    }
+
+    buffer = sqlite3_serialize(conn->db, "main", &buffer_size, 0);
+    if (!buffer) {
+        return make_error_tuple(env, "serialization_failed");
+    }
+
+    serialized = make_binary(env, buffer, buffer_size);
```
mbuhot, line 658: One alternative to copying the buffer into a BEAM binary here could be to treat the data as a resource and re-use the buffer for a subsequent deserialize call.

This felt a little safer at the expense of some copying.

c_src/sqlite3_nif.c
```diff
@@ -632,6 +632,73 @@ exqlite_transaction_status(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+        return enif_make_badarg(env);
+    }
+
+    size = serialized.size;
+    buffer = sqlite3_malloc(size);
+    if (!buffer) {
+        return make_error_tuple(env, "deserialization_failed");
+    }
+
+    rc = sqlite3_deserialize(conn->db, "main", buffer, size, size, flags);
```
mbuhot, line 694: The docs only mention `SQLITE_BUSY` as a potential error code, are there other errors I need to check for here?

c_src/sqlite3_nif.c
```diff
@@ -632,6 +632,74 @@ exqlite_transaction_status(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+    if (argc != 1) {
+        return enif_make_badarg(env);
+    }
+
+    if (!enif_get_resource(env, argv[0], connection_type, (void**)&conn)) {
+        return make_error_tuple(env, "invalid_connection");
+    }
+
+    buffer = sqlite3_serialize(conn->db, "main", &buffer_size, 0);
```
mbuhot, line 653: I've hard-coded the database name `"main"` here as that's enough for my intended usage. 
It could alternatively be taken as a parameter, and defaulted to `"main"` in `Exqlite.Sqlite3.serialize`.

warmwaffles: Let's allow this to be specified through args in `Sqlite3` and change the signature for `Sqlite3NIF` to accept a database name.

c_src/sqlite3_nif.c
warmwaffles: I think this is the safest bet. Otherwise we'd need to spin up some mutex locks to ensure that two writes don't happen to the same buffer.

c_src/sqlite3_nif.c
warmwaffles: Checkout what we do in `exqlite_execute` for returning an error code.

```c
    rc = sqlite3_exec(conn->db, (char*)bin.data, NULL, NULL, NULL);
    if (rc != SQLITE_OK) {
        return make_sqlite3_error_tuple(env, rc, conn->db);
    }
```

This will cover any cases that change in the future.

warmwaffles (changes requested):
This is close.

I've never had to use the serialize or deserialize functionality with sqlite3. What is it doing?

test/exqlite/sqlite3_test.exs
```diff
@@ -277,4 +277,23 @@ defmodule Exqlite.Sqlite3Test do
       assert :done == Sqlite3.step(conn, statement)
     end
   end
+
+  describe "serialize and deserialize" do
```
kevinlang, line 281: Nice!

mbuhot:
> I've never had to use the serialize or deserialize functionality with sqlite3. What is it doing?

`serialize` safely snapshots the state of the database to a binary. I think in some cases it's equivalent to a `File.read!` of the database file, but `serialize` ensures that there isn't a write in progress.

`deserialize` loads the serialized snapshot as an in-memory database.

The idea is that we'll be able to use this to run tests that use a sqlite database in parallel.
At the start of each test, the database will be deserialized into a temporary memory database, isolating it from other tests.

warmwaffles:
I see, thanks for clarifying. That does sound valuable.

warmwaffles (approved):
One last thing @mbuhot can you add the changes to the changelog 😄

warmwaffles:
Wewt!

PR #155, opened 2021-08-11
Adds Sqlite3.release/2 for prepared statements

### LINKED ISSUE #154: Automatically close prepared statements after execute
Right now we rely entirely on NIF destructor calling behavior which is apparently unreliable for cleaning up our queries/statements. One thing I see the `myxql` driver do is automatically close any unnamed queries after `execute`.

As part of this we can change our implementation of `handle_close` to actually close the query given.

warmwaffles:
This _should_ in theory fix #154

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -712,6 +712,35 @@ exqlite_deserialize(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     return make_atom(env, "ok");
 }
 
+static ERL_NIF_TERM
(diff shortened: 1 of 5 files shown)


warmwaffles:
Well that was weird, compilation segfaulted

warmwaffles:
I wonder if the segfault is from `enif_release_resource()`. When I run `mix test` on my local machine I do not get the segfault. I'll check if I can reproduce in a Docker container.

warmwaffles:
For anyone following along, I haven't forgotten about this. Just been busy and haven't been able to get back to this.

PR #157, opened 2021-08-25
Handle only UTC datetimes and convert to iso form without offset

LostKobrakai:
As discussed in https://github.com/elixir-sqlite/ecto_sqlite3/pull/49


warmwaffles (approved):
This looks good, can we get a short test in https://github.com/elixir-sqlite/exqlite/blob/main/test/exqlite/sqlite3_test.exs ?

PR #160, opened 2021-09-06
Support loading extensions (closes #136)

### LINKED ISSUE #136: Extension loading
I want to look into loading an extension or two from this following repository

https://github.com/nalgeon/sqlite-plus

Specifically the unicode upper and lower support. But I would love to figure out if we can support extension loading and maybe ship a way to load them. Whether it be through adding an extra dependency something like `exqlite_unicode_ext` that we provide some loading hooks.

Thoughts at @kevinlang ?

mindreframer:
As initial selection for SQLITE extensions we take https://github.com/nalgeon/sqlean with some additions in this fork (https://github.com/mindreframer/sqlean) + packaged as Hex package here: https://hex.pm/packages/ex_sqlean. 

To keep the usage for Exqlite somewhat simpler and consistent, a new module is introduced: BasicAPI. It allows simple operations without the confusion between Exqlite.Sqlite3 and Exqlite.Connection modules to execute queries and dealing with results.


mindreframer:
@warmwaffles @kevinlang @cw789  Ping?

.github/workflows/ci.yml
```diff
@@ -41,7 +42,7 @@ jobs:
             otp: "24"
           - elixir: "1.10"
             otp: "23"
-          - elixir: "1.8"
```
warmwaffles, line 44: Why the drop of 1.8?

lib/exqlite/basic_api.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Exqlite.BasicAPI do
```
warmwaffles, line 1: Is this for just testing purposes?

mix.exs
```diff
@@ -37,9 +37,10 @@ defmodule Exqlite.MixProject do
   defp deps do
     [
       {:db_connection, "~> 2.1"},
+      {:ex_sqlean, "~> 0.8.4", only: [:dev, :test]},
       {:elixir_make, "~> 0.6", runtime: false},
       {:ex_doc, "~> 0.24", only: :dev, runtime: false},
-      {:temp, "~> 0.4", only: [:test]}
+      {:temp, "~> 0.4", only: [:dev, :test]}
```
warmwaffles, line 43: Where is this used in the `dev` environment?

warmwaffles:
@mindreframer I've never actually loaded extensions before. Do you have to unload them before you close the sqlite connection?

.github/workflows/ci.yml
mindreframer: I was using :persistent_term in `ex_sqlean` and 1.8 did not have it. I was hoping that it would be present in 1.9, but then I realised that it is an OTP lib, that was introduced in 21.2. So, I could actually make 1.8 the lowest supported Elixir version.

lib/exqlite/basic_api.ex
mindreframer: Currently yes, to keep the API consistent. I feel the current API with Exqlite.Sqlite and Exqlite.Connection is somewhat confusing and there is some overlap, but none of them allow simple and straightforward usage. I could move this module to `test/support` for now, but it might be beneficial to consider providing a simple API for basic usage in the future.

mix.exs
mindreframer: I was using in the `IEx` for testing things around. Not sure if there is such a big deal between dev / test for libraries, does it make any difference if `exqlite` is included into another Elixir package / application?

mindreframer:
@warmwaffles I don't think so. Just closing SQLite connection is enough. An extension is loaded in the context of a connection and does not linger around / or leak to other connections.

.github/workflows/ci.yml
warmwaffles: I _think_ we can shift support for OTP to 21. People really shouldn't be on something that old.

lib/exqlite/basic_api.ex
warmwaffles: The `DBConnection` interface definitely confusing and unfriendly for sure. I think it's okay to introduce, though not a fan of `BasicAPI` as the name just because it's sort of redundant.

I don't know if we want to use something like `SimpleConnection` as a different name. Would that cause confusion with the `Exqlite.Connection` interface?

@kevinlang what's your opinion?

mix.exs
warmwaffles: Let's leave it, I actually remember needing to do something similarly prior.

lib/exqlite/basic_api.ex
warmwaffles: At worst, we move this to the `test/support` and once we come to a consensus we could `git mv` it out.

.github/workflows/ci.yml
mindreframer: I actually agree. There are not many practical reasons to support older OTP for a new-ish (Elixir) SQLite library, considering how easy it is to upgrade your application to newer Erlang / Elixir versions.

lib/exqlite/basic_api.ex
mindreframer: I would be fine renaming it. I just did not want to waste my mental cycles on naming yesterday, because you know... naming and cache invalidation ... hehe. We could also move those functions to the top-level module `Exqlite`, that is currently completely empty. But then again, this would encourage this API as the official one, which we might not want just yet.  

Alternative names (in no particular order and mostly )
- Exqlite.Runner
- Exqlite.API
- Exqlite.Basic
- Exqlite.Simple
- Exqlite.Easy
- Exqlite.Base
- Exqlite.Smart

warmwaffles: I think `Exqlite.Basic` would suffice.

mindreframer: Cool.

.github/workflows/ci.yml
warmwaffles: Let's bump the lowest to OTP 21 and try to support 1.8 elixir for a little while.

mindreframer:
@warmwaffles Feel free to play around and suggest some further improvements. I'm not a C programmer, so I kinda copy-pasted the NIF code together, and worked around the booleans by mapping them to 1/0 in the `sqlite3.ex` file. It might not be 100% bullet proof, but got the job done.

warmwaffles:
The C code looks completely fine.

cw789:
> @cw789 Ping?

Thanks for asking. But I think I could not help much regarding this.
OTP 21+ is also fine in my opinion.

If you really ask me, I think I would miss how I could use SQLean or other Plugins (if others exists).
A little bit of docs? ;)

mindreframer:
@cw789 Thanks for the feedback! I have added a section in the Readme with some code snippets about how to use  SQLean extensions.

mindreframer:
@warmwaffles thanks for merging!

cw789:
Thanks for making.

warmwaffles:
No problem. I was eyeballing sqlean a while back but decided against incorporating it into this library. Having it as an external is totally fine.

PR #162, opened 2021-09-09
Fix segfault on multiple connection closing (closes #161)

### LINKED ISSUE #161: [BUG] Closing an already closed connection kills the Erlang VM
Code to reproduce: 

```elixir
Interactive Elixir (1.12.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)>
nil
iex(2)> alias Exqlite.Sqlite3
Exqlite.Sqlite3
iex(3)> {:ok, conn} = Sqlite3.open(":memory:")
{:ok, #Reference<0.3373779577.804651032.117242>}
iex(4)> Sqlite3.close(conn)
:ok
iex(5)>  Sqlite3.close(conn)
fish: Job 1, 'iex -S mix' terminated by signal SIGSEGV (Address boundary error)
```

It would be nice to handle this situation more gracefully.

mindreframer:
Problem: 
- As described in https://github.com/elixir-sqlite/exqlite/issues/161, closing a connection more than once leads to a SIGSEGV termination

Solution: 
- check if conn->db is already NULL while closing and return early
+ test case for double connection closing

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -154,6 +154,11 @@ exqlite_close(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         return make_error_tuple(env, "invalid_connection");
     }
 
+    // DB is already closed, nothing to do here
+    if (conn->db == NULL) {
+        return make_atom(env, "ok");
+    }
+
     int autocommit = sqlite3_get_autocommit(conn->db);
     if (autocommit == 0) {
         rc = sqlite3_exec(conn->db, "ROLLBACK;", NULL, NULL, NULL);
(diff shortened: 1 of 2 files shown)


warmwaffles:
Thanks a ton for digging in to this.

mindreframer:
@warmwaffles  thanks for the quick hint! I was initially intimidated by this error :)

warmwaffles:
C is easy, if it's not initially written in a mess to begin with 😛

mindreframer:
Well, C is definitely easier than assembler :) `C that it’s undeniably a high-level language.`  (https://www.quora.com/Is-C-language-is-a-high-level-language-and-why). But yes, I would have to work my way towards C from the so called modern scripting languages.

PR #163, opened 2021-09-13
Add windows to CI

### LINKED ISSUE #120: Add Windows CI
Once #26 is fixed, we will want to add a Windows CI to ensure that it stays building properly.

cw789:
This is my first attempt for the Windows CI (and nearly my fist attempt to work with Github Actions).
The `extensions_test.exs` is currently failing, so I mark this as draft first.

Let me tell your thoughts?

Close #120


cw789:
@mindreframer
Hi Roman, could you probably give me a hint why the tests are failing on Windows?

warmwaffles:
@mindreframer and @cw789  this may be due to `sqlean` that was added https://github.com/mindreframer/ex_sqlean

cw789:
Definitely. And my assumption now is it's caused by 32-bit or 64-bit mismatch?

.github/workflows/ci.yml
```diff
@@ -22,13 +22,14 @@ jobs:
...
-    runs-on: ubuntu-20.04
+    runs-on: ${{ matrix.os }}
     env:
       MIX_ENV: test
-    name: Ubuntu / OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
+    name: Test ${{ matrix.os }} / OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-20.04, windows-2019]
```
warmwaffles, line 32: you should just be able to do 

```yaml
os:
  - "ubuntu-20.04"
  - "windows-2019"
```

warmwaffles:
Weird this didn't run the entire matrix of all the elixir and windows versions.

Perhaps we make a separate workflow for just windows?

warmwaffles:
> Definitely. And my assumption now is it's caused by 32-bit or 64-bit mismatch?

That's definitely a possibility.

cw789:
> Weird this didn't run the entire matrix of all the elixir, otp, ubuntu, and windows versions.

```
strategy:
      fail-fast: false
```

This would probably be needed then? But I'm still not the expert.

warmwaffles:
@cw789 I think you need to just add to the matrix include section

```yaml
    strategy:
      fail-fast: false
      matrix:
        include:
          - elixir: "1.12"
            otp: "24"
            os: ubuntu-20.04
          - elixir: "1.12"
            otp: "23"
            os: ubuntu-20.04
          - elixir: "1.11"
            otp: "24"
            os: ubuntu-20.04
          - elixir: "1.11"
            otp: "23"
            os: ubuntu-20.04
          - elixir: "1.10"
            otp: "24"
            os: ubuntu-20.04
          - elixir: "1.10"
            otp: "23"
            os: ubuntu-20.04
          - elixir: "1.8"
            otp: "21"
            os: ubuntu-20.04
          - elixir: "1.12"
            otp: "24"
            os: windows-2019
          - elixir: "1.11"
            otp: "24"
            os: windows-2019
```

warmwaffles:
I also found this tool useful for my local environment https://github.com/nektos/act don't know if windows is supported

warmwaffles:
Cool that works good. Now we are back to the original problem 😛

cw789:
Sorry for exerimenting.  Done for today.

> Perhaps we make a separate workflow for just windows?

Yeah perhaps. I also think it's not necessary to test against Windows for the complete matrix. I'll consider that tomorrow.
I don't think many people will deploy on Windows anyway.

warmwaffles:
> I don't think many people will deploy on Windows anyway.

Yea @cw789 that's why I suggested just adding two new entries to the `include` section with the OS part. Although I will be looking in to android compilation and testing as well, so I don't think it's that big of a deal to test the whole matrix against windows, linux, and possibly android.

cw789:
Added deps chaching and moved windows to `include` .
From myside this is ok now.

Still only the failing tests to solve.

warmwaffles:
@mindreframer for sqlean, could you possibly just compile the modules on demand instead of relying on prebuilt binaries?

mindreframer:
@warmwaffles Not sure if like that idea. But maybe you'll try it out with another library, that `elixir-sqlite` team can support.

I looked a bit in this, it seem that the architecture from `windows-2019` is "win-32". Just for this case I have adjusted the projects to also consider this: 
- https://github.com/mindreframer/sqlean  
   - https://github.com/mindreframer/sqlean/commit/a30dc29861e42b1be4f5f13dfcf3081fd1d6fed6)
- https://github.com/mindreframer/ex_sqlean/
   - https://github.com/mindreframer/ex_sqlean/commit/8f25406f14e16757bb83e264a8d7337b409d9cf5


Please update to `{:ex_sqlean, "~> 0.8.5"}` and it should work. 

Here is the green build on my branch: https://github.com/mindreframer/exqlite/actions/runs/1234399784


For future reference: 

` - run: mix run -e 'IO.inspect(ExSqlean.CpuInfo.fullinfo())'` -> will give you the CPU info in the build log, that helps to narrow down the issue. 

Have a nice day!

cw789:
Thanks a lot! 💛

mindreframer:
@cw789 You're welcome!

PR #164, opened 2021-09-13
Support Android compilation

dominicletz:
Hey Guys

I'm using your SQLite nif for my android project https://github.com/elixir-desktop/android-example-app but found that for Android I need to link the nif explicitly again `libm` so I added that under the `$crosscompile` variable when set to 'android'

Additionally, I've added `-Os` to optimize the binary for size and `-z global` though that will not always be needed.

I'm not sure on the reuse of `$crosscompile` since I don't know for what use case you intended it, alternatively I was thinking of using the mix std environment variable `MIX_TARGET` e.g. such as in:
`ifeq ($(MIX_TARGET), Android)`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,8 +44,13 @@ PRIV_DIR = $(MIX_APP_PATH)/priv
 LIB_NAME = $(PRIV_DIR)/sqlite3_nif.so
 
 ifneq ($(CROSSCOMPILE),)
-	LIB_CFLAGS := -shared -fPIC -fvisibility=hidden
-	SO_LDFLAGS := -Wl,-soname,libsqlite3.so.0
+	ifeq ($(CROSSCOMPILE), Android)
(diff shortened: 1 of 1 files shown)


warmwaffles:
This would be cool if you could also add it to the https://github.com/elixir-sqlite/exqlite/blob/main/.github/workflows/ci.yml as well

warmwaffles:
#163 is adding the windows CI so it could be possible that we shim this in on top of that after we get it merged.

dominicletz:
I don't think GitHub supports running Android code, does it? This issue only appears in runtime.

warmwaffles:
@dominicletz I don't see why it couldn't. I'll explore adding it to the CI workflow. But for now can you confirm if this actually builds and works in your android environment?

warmwaffles:
@dominicletz this has been deployed in `v0.7.2`

PR #167, opened 2021-10-04
Enable static compilation for iOS deployment

dominicletz:
I'm using this library to run local sqlite datbases cross platform, including iOS devices now: https://github.com/elixir-desktop/ios-example-app -- this change allows to compile the library statically and link into the app image on iOS (iOS does not have shared libraries)


warmwaffles:
If we are going to add this, I feel like we need to at least have our CI do a compilation for it 

https://github.com/yukiarrr/ios-build-action

dominicletz:
I've added a CI updated that compiles the static nif library as well, so both (shared and static) get compiled. My motivation has been iOS but the static nif can also be used e.g. for GRISP or other embedded systems there is nothing iOS specific about this change in fact.

So I think if this CI change works and compiles the static lib we should be good!

warmwaffles:
@dominicletz quick question before I merge this. When was `ERL_NIF_INIT_DECL` added? Just want to know if it'll be supported for OTP 21-24

PR #180, opened 2021-12-07
Support mkdir -p behavior when DB file does not exist

jjcarstens:
Currently, if you supply a DB path that includes directories which do not exist, sqlite3 will not attempt to create them and the open will crash.

This adds a quick attempt to `mkdir` on the db path to ensure all intermediary directories exist before attempting to create the DB file for cases that `SQLITE_OPEN_CREATE` flag is enabled and would expect the file to be created when non-existant

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -389,7 +389,8 @@ defmodule Exqlite.Connection do
   end
 
   defp do_connect(path, options) do
-    with {:ok, db} <- Sqlite3.open(path),
+    with :ok <- mkdir_p(path),
+         {:ok, db} <- Sqlite3.open(path),
          :ok <- set_journal_mode(db, options),
          :ok <- set_temp_store(db, options),
          :ok <- set_synchronous(db, options),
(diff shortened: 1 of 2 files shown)


warmwaffles:
Why can't this be solved at the elixir level? I'd rather stay away from modifying `c_src/sqlite3.c` as that is the sqlite amalgamation file.

jjcarstens:
@warmwaffles You're totally right - In fact, @fhunleth and I were just talking about this. I just jumped the gun to C for who-knows-why 🤦 

Sit tight - Better change coming up soon

warmwaffles:
@jjcarstens it's not a big deal, I admire the effort to dig into the sqlite3.c file. That thing is a pain to load in the text editor.

jjcarstens:
Ok @warmwaffles - Much easier in Elixir 🤣

warmwaffles:
Much cleaner too!

warmwaffles:
@jjcarstens I cut a release https://github.com/elixir-sqlite/exqlite/releases/tag/v0.8.3

jjcarstens:
So fast! Thanks @warmwaffles ❤️

PR #182, opened 2021-12-14
Fix CI

cw789:
Some CI actions will fail due to Hex v1.0.0.

Also add Elixir v1.13.


cw789:
Ok, an upstream fix is available.
So with Hex v1.0.1 the combination of Elixir 1.11.4 / OTP 24 probably could be re enabled.

warmwaffles:
Thanks @cw789

PR #186, opened 2022-01-21
Allow using system-wide or custom sqlite3 installation

### LINKED ISSUE #178: Allow linking with system library
What I would like to do is provide the consumer the possibility to link against the sqlite3 binary from the OS or to compile it directly.

I'm thinking maybe doing `EXQLITE_USE_SYSTEM=1` as the environment variable to use for that.

Another solution would be to completely toss the code and make it a requirement that the consuming system has the library installed.

ConnorRigby:
This allows using engines built by other projects. For
example with https://www.zetetic.net/sqlcipher or
https://www.sqlite.org/see/doc/trunk/www/readme.wiki

Fixes #178

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,15 +16,32 @@
 # LDFLAGS       linker flags for linking all binaries
 #
 
-SRC = $(wildcard c_src/*.c)
-HEADERS = $(wildcard c_src/*.h)
+SRC = c_src/sqlite3_nif.c
+HEADERS = c_src/utf8.h
+
+ifeq ($(EXQLITE_USE_SYSTEM),)
+	SRC += c_src/sqlite3.c
+	HEADERS += c_src/sqlite3.h c_src/sqlite3ext.h
+	CFLAGS += -Ic_src
+else
+	ifneq ($(EXQLITE_SYSTEM_CFLAGS),)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,15 +16,32 @@
 # LDFLAGS       linker flags for linking all binaries
 #
 
-SRC = $(wildcard c_src/*.c)
-HEADERS = $(wildcard c_src/*.h)
+SRC = c_src/sqlite3_nif.c
+HEADERS = c_src/utf8.h
+
+ifeq ($(ESQLITE_USE_SYSTEM),)
```
warmwaffles, line 22: We'll want to change this to `EXQLITE_USE_SYSTEM`

ConnorRigby: oops, muscle memory kicked in. Updating..

Makefile
```diff
@@ -126,5 +146,3 @@ clean:
 
 .PHONY: all clean
 
-# Don't echo commands unless the caller exports "V=1"
-${V}.SILENT:
```
warmwaffles, line 130: One last thing, we added this in to clean up the compilation output. Was introduced here https://github.com/elixir-sqlite/exqlite/pull/181

ConnorRigby: oops sorry about that one.. just a sec

PR #193, opened 2022-02-22
Define custom memory allocation definitions

warmwaffles:
What my aim here is to expose the amount of memory being utilized by an erlang vm process that is currently running.

I am opening this up for comment and am probably going to add in some benchmarks and some load testing scripts to go along with this.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -12,6 +12,7 @@
 
 static ErlNifResourceType* connection_type = NULL;
 static ErlNifResourceType* statement_type  = NULL;
+static sqlite3_mem_methods default_alloc_methods = {0};
 
 typedef struct connection
 {
@@ -871,6 +872,7 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
       exqlite_mem_shutdown,
       0};
 
+    sqlite3_config(SQLITE_CONFIG_GETMALLOC, &default_alloc_methods);
     sqlite3_config(SQLITE_CONFIG_MALLOC, &methods);
 
     connection_type = enif_open_resource_type(
@@ -903,7 +905,7 @@ on_unload(ErlNifEnv* caller_env, void* priv_data)
 {
     assert(env);
 
-    sqlite3_config(SQLITE_CONFIG_MALLOC, NULL);
+    sqlite3_config(SQLITE_CONFIG_MALLOC, &default_alloc_methods);
 }
 
 //
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -23,6 +23,83 @@ typedef struct statement
     sqlite3_stmt* statement;
 } statement_t;
 
+static void*
+exqlite_malloc(int bytes)
+{
+    assert(bytes > 0);
+
+    int* p = enif_alloc(bytes + sizeof(int));
+    if (p) {
+        p[0] = bytes;
+        p++;
+    }
+
+    return p;
+}
+
+static void
+exqlite_free(void* prior)
+{
+    if (!prior) {
+        return;
+    }
+
+    int* p = prior;
+
+    // Shift the pointer back to free the proper block of data
+    p--;
+
+    return enif_free(p);
+}
+
+static void*
+exqlite_realloc(void* prior, int bytes)
+{
+    assert(prior);
+    assert(bytes > 0);
+
+    int* p = prior;
+    p--;
+
+    p = enif_realloc(p, bytes + sizeof(int));
+    if (p) {
+        p[0] = bytes;
+        p++;
+    }
+
+    return p;
+}
+
+static int
+exqlite_mem_size(void* prior)
+{
+    if (!prior) {
+        return 0;
+    }
+
+    int* p = prior;
+    p--;
+
+    return p[0];
+}
+
+static int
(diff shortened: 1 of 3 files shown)


laszlohegedus:
Hey Matthew, our app fails with segmentation fault with these changes. I'll look into it, but can you also check what may cause it?

warmwaffles:
> Hey Matthew, our app fails with segmentation fault with these changes. I'll look into it, but can you also check what may cause it?

This is good to know. I'll need to do some more load testing with it.

warmwaffles:
@laszlohegedus could you provide some information on what may have caused the segfault?

laszlohegedus:
> @laszlohegedus could you provide some information on what may have caused the segfault?

Sure, it actually happens on startup. I want to check it with Valgrind and see if it tells us why. We're also using libbtc_ex. I wonder if they collide with exqlite. Maybe I'll just try the two in a dummy project.

laszlohegedus:
@warmwaffles Well, it was a pain indeed. I managed to build the beam with valgrind and even "hack" the Elixir executable to use `cerl`, but it didn't want to work properly with our application, so I gave up.

What we can tell now is that we work with exqlite master, but not these changes. I'll dig into the code to see where it fails. At a first glance with my rarely used knowledge of C, the changes look good.

laszlohegedus:
@warmwaffles I think I found the issue. It's in the `on_unload` function, when the memory allocator functions are set to `NULL`. We need to revert to the defaults there. I couldn't find a way to access that struct from outside of `sqlite.c`, but we can use `SQLITE_CONFIG_GETMALLOC` before we set our functions and store the defaults in a global variable. I'll add some comments to the code.

c_src/sqlite3_nif.c
```diff
@@ -785,6 +861,18 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
...
+      exqlite_malloc,
+      exqlite_free,
+      exqlite_realloc,
+      exqlite_mem_size,
+      exqlite_mem_round_up,
+      exqlite_mem_init,
+      exqlite_mem_shutdown,
+      0};
+
+    sqlite3_config(SQLITE_CONFIG_MALLOC, &methods);
```
laszlohegedus, line 874: Before we do this I suggest adding a global variable in the top of the file:
```
static sqlite3_mem_methods default_alloc_methods;
```

then
```suggestion
    sqlite3_config(SQLITE_CONFIG_GETMALLOC, &default_alloc_methods);
    sqlite3_config(SQLITE_CONFIG_MALLOC, &methods);
```

c_src/sqlite3_nif.c
```diff
@@ -810,6 +898,13 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
     return 0;
 }
 
+static void
+on_unload(ErlNifEnv* caller_env, void* priv_data)
+{
+    assert(env);
+
+    sqlite3_config(SQLITE_CONFIG_MALLOC, NULL);
```
laszlohegedus, line 906:
```suggestion
    sqlite3_config(SQLITE_CONFIG_MALLOC, &default_alloc_methods);
```

c_src/sqlite3_nif.c
```diff
@@ -810,6 +898,13 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
     return 0;
 }
 
+static void
+on_unload(ErlNifEnv* caller_env, void* priv_data)
```
laszlohegedus, line 902:
```suggestion
on_unload(ErlNifEnv* env, void* priv_data)
```

warmwaffles: We are going to leave this as `caller_env` as that is what it is defined as within the erl_nif documentation. https://www.erlang.org/doc/man/erl_nif.html#initialization

c_src/sqlite3_nif.c
warmwaffles: It could just be better to not do this at all.

warmwaffles:
@laszlohegedus see the latest commit, I lifted the struct method definitions to a static global context, and removed the `on_unload` definition. I wanted to try and be a "good citizen" and clear what was set on unload, but it probably doesn't matter in the grand scheme of things.

warmwaffles:
> The `SQLITE_CONFIG_MALLOC` option takes a single argument which is a pointer to an instance of the sqlite3_mem_methods structure. The argument specifies alternative low-level memory allocation routines to be used in place of the memory allocation routines built into SQLite. SQLite makes its own private copy of the content of the sqlite3_mem_methods structure before the sqlite3_config() call returns.
>
> The `SQLITE_CONFIG_GETMALLOC` option takes a single argument which is a pointer to an instance of the sqlite3_mem_methods structure. The sqlite3_mem_methods structure is filled with the currently defined memory allocation routines. This option can be used to overload the default memory allocation routines with a wrapper that simulations memory allocation failure or tracks memory usage, for example. 

I'm unsure which one is the one we have to set.

c_src/sqlite3_nif.c
warmwaffles: Implemented! Thanks

c_src/sqlite3_nif.c
warmwaffles: Implemented the change.

laszlohegedus (approved):
Looks great. Thank you.

PR #194, opened 2022-03-01
fixed path to Visual Studio 2022 vcvars64.bat

mellkior:
Visual Studio 2022 is 64 bit, unlike previous versions of Visual Studio, so by default it should reside under `C:\Program Files\` rather than `C:\Program Files (x86)`. I updated the paths to vcvars64.bat to what it is on my windows 10 machine with Visual Studio 2022 version 17.1.0.

### DIFF
--- a/guides/windows.md
+++ b/guides/windows.md
@@ -50,7 +50,7 @@ mix test
 installed in its default installation path.
 
 ```powershell
-cmd /k "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC\Auxiliary\Build\vcvars64.bat"
+cmd /k "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
 ```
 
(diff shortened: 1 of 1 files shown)


warmwaffles:
Weird, I wonder why the checks aren't running

warmwaffles (approved):
This looks good to me, but I just need to figure out why the actions are not working.

cw789:
I can't confirm this change is correct for the current guide.
I've written the guide for Visual Studio **Build Tools** only - not for a full Visual Studio installation.

cw789:
Not really interested, at the moment. Just noting my concerns.
I mean if you follow the whole guide now with installing Build Tools instead of the gigantic Visual Studio - I think the path is wrong now.

But I've not done a fresh installation of the Build Tools myself. At least for my upgraded version (17.1) the former was still correct.

warmwaffles:
I'll revert this change and ship it with the upgrade to 3.38 in #195

PR #197, opened 2022-04-06
Remove semicolons in README

LostKobrakai:


warmwaffles:
Hahah @LostKobrakai I must have written these just after mucking around in the C code. Old habbits die hard. 😆

Issue #196, opened 2022-03-29
Unable to handle BIGINT on 32 Bit arm


alanj853:
Hi,

I am currently working on a custom armv7 board (similar to a raspberry pi), and am having trouble upgrading to using exqlite. I was using [esqlite](https://github.com/mmzeeman/esqlite), and everything was working fine, but due to the fact that https://github.com/elixir-sqlite/sqlitex and https://github.com/elixir-sqlite/sqlite_ecto2 are no longer being maintained and needed to use ecto 2, I am trying to switch over to the [exqlite](https://github.com/elixir-sqlite/exqlite), [ecto_sqlite3](https://github.com/elixir-sqlite/ecto_sqlite3) combination where I can also get ecto 3.

The issue in exqlite is failing to handle integers greater than 32 bit in size during a schema migration, although esqlite can handle them. For example, for the integer 20180801160101, -1250163803 is being returned when doing a multistep.

The ecto query is
```
#Ecto.Query<from s0 in "schema_migrations", select: type(s0.version, :integer)>
```
Unfortunately, I am unable to upload the sqlite3 file I am using due to GitHub restrictions.

Some things to note:
- This works fine on 64 bit machines
- Both esqlite and exqlite are using sqlite v3.38 (as in the latest of both repositories)
- I have not changed the schema_migrations schema. It is this one that is getting used: https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/migration/schema_migration.ex#L10

**Software Versions:**
- Elixir 1.12.3-otp-24
- Erlang 24.1.3
- Linux Kernel 5.4.66
- exqlite 0.10.2
- sqlite 3.38
- ecto_sql - 3.380-dev
- ecto 3.7.0
- ecto_sqlite3 - 0.7.4

warmwaffles:
I'll take a look at this soon. I don't have easy access to 32 bit arm at the moment.

Could you do me a favor and blow away your `_build` and then do `V=1 mix deps.compile exqlite`

warmwaffles:
When I get some time this evening I'll try to recreate this issue. What would be really helpful is setting up a dummy project for me to check out and poke around in.

Can you also take a look and downgrade exqlite to `0.9.3`? In the change from [0.9.3 to 0.10.2](https://github.com/elixir-sqlite/exqlite/compare/v0.9.3...HEAD) I added a custom allocator and I am wondering if that is causing any issues. I doubt it but I just want to check all of my bases.

[...]

PR #198, opened 2022-04-07
Adding ERL_CFLAGS

alanj853:
Is possibly required if doing a cross-compilation.

Addresses https://github.com/elixir-sqlite/exqlite/issues/196


warmwaffles:
@alanj853 we'll need to add this same thing to the windows makefile as well.

alanj853:
@warmwaffles , that should be done now.

warmwaffles:
🥳

Issue #199, opened 2022-04-08
Big result queries are slow compared to SQLite3 CLI or other clients 


SebastianSzturo:
I've noticed that large queries are significantly slower via `ecto_sqlite3` than through the `sqlite3` CLI when using Ecto or `Exqlite.Connection` (not with `Exlite.Sqlite3`)

Here is the example data of ~650k rows: [dev.db.zip](https://github.com/elixir-sqlite/ecto_sqlite3/files/8449228/dev.db.zip)

Via SQLite3 CLI or any other SQLite3 GUI (with wal mode enabled):
```
sqlite> .timer on
sqlite> SELECT transaction_date, transaction_amount FROM transactions;
Run Time: real 2.680 user 0.423684 sys 0.323269
sqlite>
```
 
 With `Exqlite.Connection`: **21.98 s**
 ```elixir
     {:ok, db} =
      DBConnection.start_link(Exqlite.Connection,
        database: "./dev.db",
        journal_mode: :wal,
        cache_size: -64000,
        temp_store: :memory
      )

    {:ok, _, result} =
      DBConnection.execute(
        db,
        %Exqlite.Query{
          statement: "SELECT transaction_date, transaction_amount FROM transactions"
        },
        []
      )
 ```
 ```
 Name                         ips        average  deviation         median         99th %
exqlite_connection     0.00005 K        21.98 s     ±0.00%        21.98 s        21.98 s
```

Same with Ecto : **21520.2ms**
```elixir
Ecto.Adapters.SQL.query(
  Repo,
  "SELECT transaction_date, transaction_amount FROM transactions"
)
```
```
[debug] QUERY OK db=21520.2ms decode=1.4ms idle=106.7ms
SELECT transaction_date, transaction_amount FROM transactions []
```

Any idea what this could be caused by and how to debug it further?

SebastianSzturo:
Here is a full benchmark example of the performance difference with a `limit 10000` to make the performance difference clearer without having the client timeout in the connection example:

```elixir
Benchee.run(%{
  "exqlite_direct" => fn ->
    {:ok, conn} = Exqlite.Sqlite3.open("./dev.db")

    {:ok, statement} =
      Exqlite.Sqlite3.prepare(
        conn,
        "SELECT transaction_date, transaction_amount FROM transactions limit 10000"
      )
  end,
  "exqlite_connection" => fn ->
    {:ok, db} =
      DBConnection.start_link(Exqlite.Connection,
[...]

PR #200, opened 2022-04-10
Improve performance of `fetch_all/4`

SebastianSzturo:
In Erlang/Elixir appending to a list requires to traverse the whole list first to find it's end which can be quite slow. The recommendation is to prepend the list and reverse to improve performance:
* https://www.erlang.org/doc/efficiency_guide/myths.html#myth--operator--++--is-always-bad
* https://www.wyeworks.com/blog/2019/03/01/to-use-or-not-to-use-the-++-operator-in-elixir/
* https://stackoverflow.com/questions/3232786/how-to-concat-lists-in-erlang-without-creating-nested-lists

--

This PR is trying to use this method to improve performance of `fetch_all/4` (https://github.com/elixir-sqlite/exqlite/issues/199)
* There is an additional improvement that could be made by breaking the existing api for `multi_step` or having an additional private [`multi_step/3`](https://github.com/elixir-sqlite/exqlite/blob/main/lib/exqlite/sqlite3.ex#L100) that does not `Enum.reverse()` rows if we plan to accumulate the result. (We currently reverse additional rows twice)
* Maybe we should move the list-building to C in the long-term, which might come with other problems by adding time to the NIF.

### Benchmarks:
100 records: **1.05x slower**
```
Name                    ips        average  deviation         median         99th %
old_fetch_all        3.91 K      255.84 μs    ±77.49%         257 μs      517.18 μs
new_fetch_all        3.71 K      269.40 μs    ±85.54%      262.92 μs      527.12 μs

Comparison: 
old_fetch_all        3.91 K
new_fetch_all        3.71 K - 1.05x slower +13.57 μs
```

1.000 records: **1.11x  faster**
```
Name                    ips        average  deviation         median         99th %
new_fetch_all        1.00 K        1.00 ms    ±23.28%        0.97 ms        1.83 ms
old_fetch_all        0.90 K        1.11 ms    ±18.62%        1.09 ms        1.72 ms

Comparison: 
new_fetch_all        1.00 K
old_fetch_all        0.90 K - 1.11x slower +0.111 ms
```

10.000 records: **2.10x faster**
```
Name                    ips        average  deviation         median         99th %
new_fetch_all        159.90        6.25 ms    ±14.28%        6.04 ms        9.31 ms
old_fetch_all         76.04       13.15 ms     ±9.02%       13.05 ms       14.71 ms

Comparison: 
new_fetch_all        159.90
old_fetch_all         76.04 - 2.10x slower +6.90 ms
```

100.000 records: **5.23x faster**
```
Name                    ips        average  deviation         median         99th %
new_fetch_all         13.22       75.63 ms     ±7.55%       75.59 ms       91.79 ms
old_fetch_all          2.53      395.56 ms     ±0.84%      395.35 ms      400.38 ms

Comparison: 
new_fetch_all         13.22
old_fetch_all          2.53 - 5.23x slower +319.93 ms
```

1.000.000 records: **73.99x faster**
```
Name                    ips        average  deviation         median         99th %
new_fetch_all          1.38         0.72 s     ±3.95%         0.72 s         0.77 s
old_fetch_all        0.0187        53.46 s     ±0.00%        53.46 s        53.46 s

Comparison: 
new_fetch_all          1.38
old_fetch_all        0.0187 - 73.99x slower +52.74 s
```

<details>
<summary>
Benchmark Code
</summary>

```elixir

path = Temp.path!()
File.rm(path)

{:ok, db} = Exqlite.Sqlite3.open(path)

:ok =
  Exqlite.Sqlite3.execute(db, "create table users (id integer primary key, name text)")

IO.puts("Inserting rows")

Enum.each(1..1000, fn i ->
  IO.puts("Inserting #{i}...")

  :ok =
    Exqlite.Sqlite3.execute(
      db,
      "insert into users (id, name) values (#{i}, 'User-#{i}')"
    )
end)

:ok = Exqlite.Sqlite3.close(db)
IO.puts("inserting done")

Benchee.run(%{
  "old_fetch_all" => fn ->
    {:ok, db} = Exqlite.Sqlite3.open(path)

    {:ok, statement} =
      Exqlite.Sqlite3.prepare(
        db,
        "select * from users"
      )

    {:ok, _rows} = Exqlite.Sqlite3.old_fetch_all(db, statement)
  end,
  "new_fetch_all" => fn ->
    {:ok, db} = Exqlite.Sqlite3.open(path)

    {:ok, statement} =
      Exqlite.Sqlite3.prepare(
        db,
        "select * from users"
      )

    {:ok, _rows} = Exqlite.Sqlite3.fetch_all(db, statement)
  end
})

File.rm(path)
```
</details>

### DIFF
--- a/lib/exqlite/sqlite3.ex
+++ b/lib/exqlite/sqlite3.ex
@@ -135,10 +135,13 @@ defmodule Exqlite.Sqlite3 do
   defp fetch_all(conn, statement, chunk_size, accum) do
     case multi_step(conn, statement, chunk_size) do
       {:done, rows} ->
-        {:ok, accum ++ rows}
+        case accum do
+          [] -> {:ok, rows}
+          accum -> {:ok, (rows ++ accum) |> Enum.reverse()}
+        end
 
       {:rows, rows} ->
-        fetch_all(conn, statement, chunk_size, accum ++ rows)
+        fetch_all(conn, statement, chunk_size, (rows |> Enum.reverse()) ++ accum)
 
       {:error, reason} ->
         {:error, reason}
--- a/test/exqlite/connection_test.exs
+++ b/test/exqlite/connection_test.exs
@@ -157,6 +157,43 @@ defmodule Exqlite.ConnectionTest do
 
       File.rm(path)
     end
+
+    test "returns timely and in order for big data sets" do
+      path = Temp.path!()
+
+      {:ok, db} = Sqlite3.open(path)
+
+      :ok =
+        Sqlite3.execute(db, "create table users (id integer primary key, name text)")
+
+      Enum.each(1..10_000, fn i ->
+        :ok =
+          Sqlite3.execute(db, "insert into users (id, name) values (#{i}, 'User-#{i}')")
+      end)
+
+      :ok = Exqlite.Sqlite3.close(db)
+
+      {:ok, conn} = Connection.connect(database: path)
+
+      {:ok, _query, result, _conn} =
+        Connection.handle_execute(
+          %Exqlite.Query{
+            statement: "SELECT * FROM users"
+          },
+          [],
+          [timeout: 1],
+          conn
+        )
+
+      assert result.command == :execute
+      assert length(result.rows) == 10_000
+
+      Enum.with_index(result.rows, fn row, i ->
+        assert row == [i + 1, "User-#{i + 1}"]
+      end)
+
+      File.rm(path)
+    end
   end
 
   describe ".handle_prepare/3" do


lib/exqlite/sqlite3.ex
```diff
@@ -135,10 +135,13 @@ defmodule Exqlite.Sqlite3 do
...
       {:done, rows} ->
-        {:ok, accum ++ rows}
+        case accum do
+          [] -> {:ok, rows}
+          accum -> {:ok, (rows ++ accum) |> Enum.reverse()}
+        end
 
       {:rows, rows} ->
-        fetch_all(conn, statement, chunk_size, accum ++ rows)
+        fetch_all(conn, statement, chunk_size, (rows |> Enum.reverse()) ++ accum)
```
SebastianSzturo, line 144: We need to reverse the records inside the list before adding them, since we have a nested list of `rows` e.g.`[[1, "Bob"], [2, "Peter"]]`

warmwaffles (approved):
This is actually something I have been needing to get around to taking care of since implemented. I knew that using `++` with two lists was not performant. Originally I did not use `multi_step` and just stepped through the results, and prepended to the list like you are supposed to do and then reverse it.

Problem was, the overhead to calling `step` multiple times in a tight loop was so slow versus just batching the results into chunks.

PR #203, opened 2022-05-05
Adds similar interface as Postgrex offers

warmwaffles:


aleDsz:
I think this is all good, I've tried here in my KinoDB implementation and it works 🎉

aleDsz (approved):

Issue #149, opened 2021-05-03
Use sqlite3_open_v2


warmwaffles:
We can specify a read only flag, which would be extra helpful for an ecto repo that has a flag set to `readonly`. This would ensure that if a bug is introduced upstream we'd have the lower level block the write.

https://www.sqlite.org/c3ref/open.html

PR #211, opened 2022-08-07
Use sqlite3_open_v2

greg-rychlewski:
Closes https://github.com/elixir-sqlite/exqlite/issues/149

- Uses `sqlite3_open_v2` instead of `sqlite3_open` inside of the nif.
- Exposes an optional second argument to `Exqlite.Sqlite3.open` to specify options, with `:mode` being one of them. `:mode`  defaults to `:readwrite` (the existing behaviour, including creating the db if it doesn't exist) and also allows `:readonly`. There are a bunch of flags that can be used for open_v2 so that's why I made this `opts` instead of a single argument for mode.

If merged I can make a change to the ecto adapter to allow it to utilize the new parameter to `Exqlite.Sqlite3.open`

### DIFF
--- a/lib/exqlite/sqlite3.ex
+++ b/lib/exqlite/sqlite3.ex
@@ -12,12 +12,14 @@ defmodule Exqlite.Sqlite3 do
   # Need to figure out if we can just stream results where we use this
   # module as a sink.
 
-  alias Exqlite.{Flags, Sqlite3NIF}
+  alias Exqlite.Flags
+  alias Exqlite.Sqlite3NIF
 
   @type db() :: reference()
   @type statement() :: reference()
   @type reason() :: atom() | String.t()
   @type row() :: list()
+  @type open_opt :: {:mode, :readwrite | :readonly}
 
   @doc """
   Opens a new sqlite database at the Path provided.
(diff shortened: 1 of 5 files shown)


lib/exqlite/sqlite3.ex
```diff
@@ -22,10 +22,30 @@ defmodule Exqlite.Sqlite3 do
...
+
+  ## Options
+
+    * `:mode` - use `:readwrite` to open the database for reading and writing
+      or `:readonly` to open it in read-only mode. `:readwrite` will also create
+      the database if it doesn't already exist. Defaults to `:readwrite`.
   """
   @spec open(String.t()) :: {:ok, db()} | {:error, reason()}
-  def open(path), do: Sqlite3NIF.open(String.to_charlist(path))
+  def open(path, opts \\ []) do
```
warmwaffles, line 34: We should add an `open_opt` type

```elixir
@type open_opt :: {:mode, :readwrite | :readonly}
@spec open(String.t(), [open_opt()]) :: {:ok, db()} | {:error, reason()}
```

warmwaffles (approved):

PR #213, opened 2022-08-22
Migrating static nif code for OTP25

dominicletz:
With OTP25 the Erlang team has introduced an official macro to support Elixir statically linked nifs `STATIC_ERLANG_NIF_LIBNAME`. This commit switches the compilation unit to use `STATIC_ERLANG_NIF_LIBNAME` instead of the previous macro override hack.


dominicletz:
I'm using this change for https://github.com/elixir-desktop/ios-example-app - which packages exqlite to run on iOS. This now runs on stock OTP25 vs. previously only on a custom branch of OTP24

PR #214, opened 2022-08-24
Fix android build adding `-lm` to LD_FLAGS

dominicletz:
This got somewhere lost between 0.7.5 and the current master. But on android we need to explicitly link against libm to resolve `acos` and other math symbols


warmwaffles:
I don't remember why, but I suspect it may be due to windows.

PR #216, opened 2022-09-29
Update sqlite3 to 3.39.4

ruslandoga:
Just got an email about this one.

> SQLite version 3.39.4 is now available on the website:
> 
>  https://sqlite.org/index.html
>  https://sqlite.org/download.html
> 
> Version 3.39.4 is a minimal patch against the prior release that addresses issues found since the prior release.  In particular, a potential vulnerability in the FTS3 extension has been fixed, so this should be considered a security update.
> 
> In order to exploit the vulnerability, an attacker must have full SQL access and must be able to construct a corrupt database with over 2GB of FTS3 content.  The problem arises from a 32-bit signed integer overflow.


warmwaffles:
Hah, should be an easy bump.

Issue #190, opened 2022-01-28
Segfault when exceeding DBConnection timeout


LostKobrakai:
I was debugging a sudden reboot on our nerves system with a expensive query. I had the memory usage and OOM in mind, but today I found out that raising the timeout actually makes the query succeed (no reboot). Making the timout be exceeded on my local machine I could reproduce the issue and even get the seg fault message (see below). I don't think the query matters for as long as the timeout is exceeded.

```
iex(6)> Task.start(fn -> Repo.checkout(fn -> Ecto.Adapters.SQL.query!(Repo, query) end, timeout: :timer.seconds(1)) end)
{:ok, #PID<0.2359.0>}
iex(7)> 11:24:58.427 [error] Exqlite.Connection (#PID<0.2164.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.2359.0> timed out because it queued and checked out the connection for longer than 1000ms

#PID<0.2359.0> was at location:

    (exqlite 0.8.6) lib/exqlite/sqlite3.ex:91: Exqlite.Sqlite3.multi_step/3
    (exqlite 0.8.6) lib/exqlite/sqlite3.ex:138: Exqlite.Sqlite3.fetch_all/4
    (exqlite 0.8.6) lib/exqlite/connection.ex:529: Exqlite.Connection.get_rows/2
    (exqlite 0.8.6) lib/exqlite/connection.ex:475: Exqlite.Connection.execute/4
    (db_connection 2.4.1) lib/db_connection/holder.ex:354: DBConnection.Holder.holder_apply/4
    (db_connection 2.4.1) lib/db_connection.ex:1333: DBConnection.run_execute/5
    (db_connection 2.4.1) lib/db_connection.ex:650: DBConnection.execute/4

11:25:00.283 [debug] …
[os_mon] cpu supervisor port (cpu_sup): Erlang has closed
[os_mon] memory supervisor port (memsup): Erlang has closed
[1]    37300 segmentation fault  iex -S mix
```

LostKobrakai:
I just saw this one, which is probably related: https://github.com/elixir-sqlite/exqlite/issues/127

SebastianSzturo:
I am still seeing the segfault described above in my app.

```
19:21:45.978 [error] Exqlite.Connection (#PID<0.2619.0>) disconnected: ** (DBConnection.ConnectionError) client #PID<0.2866.0> timed out because it queued and checked out the connection for longer than 15000m
[...]

PR #218, opened 2022-10-25
Fix segfault when database connections timeout

akash-akya:
### Hypothesis
the issue seems to be that there is a race condition between `exqlite_close` and `exqlite_prepare`. The issue happens when we execute `sqlite3_prepare_v3` on a connection after we close the connection with `sqlite3_close_v2`. 

`sqlite3_close_v2` does handle previously prepared statements, but here we are preparing the statement in-parallel while we are closing the connection. Which is causing occasional [SQLITE_MISUSE](https://www.sqlite.org/rescode.html#misuse).

How can we reach this state?

**Given**
* `exqlite_close` is executed by connection process
* while `handle_prepare` and other callbacks are called in the client process
* a NIF call can not be interrupted/cancelled [once started](https://www.erlang.org/doc/man/erl_nif.html), irrespective of client side process timeout, or process kill.

## Scenario 

1. client checkout a connection and makes a query
2. client makes `exqlite_prepare` NIF call but fail to complete within connection timeout duration. It 1ms in the test case. This NIF call is still running or scheduled to run even after the timeout.
3. due to connection timeout, connection-process receives timeout and initiate connection close (`disconnect` callback)
4. connection-process calls `exqlite_close` NIF 

2 & 4 are run in-parallel leading to race-condition mentioned previously causing segfault.

## Fix

I tried to fix the race-condition using mutex, and sofar haven't seen a single failure. 

This is PoC, I guess we should handle multiple other cases as well, not just the one between close & prepare if this is the issue.

Should fix https://github.com/elixir-sqlite/exqlite/issues/190

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -22,6 +22,7 @@ static sqlite3_mem_methods default_alloc_methods = {0};
 typedef struct connection
 {
     sqlite3* db;
+    ErlNifMutex* mutex;
 } connection_t;
 
 typedef struct statement
@@ -219,6 +220,12 @@ exqlite_open(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         return make_error_tuple(env, "database_open_failed");
     }
 
+    conn->mutex = enif_mutex_create("exqlite:connection");
+    if (conn->mutex == NULL) {
+        enif_release_resource(conn);
+        return make_error_tuple(env, "failed_to_create_mutex");
+    }
+
     sqlite3_busy_timeout(conn->db, 2000);
 
     result = enif_make_resource(env, conn);
@@ -256,6 +263,11 @@ exqlite_close(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         }
     }
 
+    // close connection in critical section to avoid race-condition
+    // cases. Cases such as query timeout and connection pooling
+    // attempting to close the connection
+    enif_mutex_lock(conn->mutex);
+
     // note: _v2 may not fully close the connection, hence why we check if
     // any transaction is open above, to make sure other connections aren't blocked.
     // v1 is guaranteed to close or error, but will return error if any
@@ -263,10 +275,12 @@ exqlite_close(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     // to later run to clean those up
     rc = sqlite3_close_v2(conn->db);
     if (rc != SQLITE_OK) {
+        enif_mutex_unlock(conn->mutex);
         return make_sqlite3_error_tuple(env, rc, conn->db);
     }
 
     conn->db = NULL;
+    enif_mutex_unlock(conn->mutex);
 
     return make_atom(env, "ok");
 }
@@ -357,11 +371,21 @@ exqlite_prepare(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     if (!statement) {
         return make_error_tuple(env, "out_of_memory");
     }
+    statement->statement = NULL;
 
     enif_keep_resource(conn);
     statement->conn = conn;
 
+    // ensure connection is not getting closed by parallel thread
+    enif_mutex_lock(conn->mutex);
+    if (conn->db == NULL) {
+        enif_mutex_unlock(conn->mutex);
+        enif_release_resource(statement);
+        return make_error_tuple(env, "connection closed");
+    }
     rc = sqlite3_prepare_v3(conn->db, (char*)bin.data, bin.size, 0, &statement->statement, NULL);
+    enif_mutex_unlock(conn->mutex);
+
     if (rc != SQLITE_OK) {
         enif_release_resource(statement);
         return make_sqlite3_error_tuple(env, rc, conn->db);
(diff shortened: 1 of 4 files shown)


c_src/sqlite3_nif.c
```diff
@@ -219,6 +220,12 @@ exqlite_open(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         return make_error_tuple(env, "database_open_failed");
     }
 
+    conn->mutex = enif_mutex_create("exqlite:connection");
+    if (conn->mutex == NULL) {
+        enif_release_resource(conn);
+        return make_error_tuple(env, "failed to create mutex");
```
warmwaffles, line 226: Since this gets turned into an atom. `{:error, :failed_to_create_mutex}`

```suggestion
        return make_error_tuple(env, "failed_to_create_mutex");
```

warmwaffles:
@akash-akya can you point the branch at `main` and cherry-pick in the test I made? I'd rather you take the credit for solving this.

Also thanks for fixing the CI goof up.

c_src/sqlite3_nif.c
akash-akya: fixed

akash-akya:
Hey @warmwaffles, made changes as suggested.

Disabling the segfault test by default since it is very slow. Let me know if you want to enable it.

Also, I am not sure if this is the right approach to solve the issue. Since we to properly fix it I think we have to wrap every callback inside mutex not just prepare.

How about this? When connection timeout, we just roll back the trx, but we **don't** close the connection at that point of time. 
Since a timed out connection will be eventually garbage collected and since at-max there can only be one parallel command, we can just let `connection_type_destructor` function close the connection at the end after the parallel command finishes. That way we don't need any mutex & wrapping, wdys?

warmwaffles:
> Disabling the segfault test by default since it is very slow. Let me know if you want to enable it.

Let's leave it disabled. I'll move it to a regression / integration level.

> Alternatively, how about this? When connection timeout, we just roll back the trx, but we don't close the connection at that point in time.

That seems like a good solution versus the mutex manipulations. But what I fear is that you can still segfault this manually using Exqlite directly. Is that okay if it segfaults if the library is used incorrectly?

greg-rychlewski:
Can you put a couple flags in the connection, something like `in_use` and `marked_for_closure`. If not in use then the timeout closes the connection. If in use then set `marked_for_closure` to true and the command will close the connection after it's done.

akash-akya:
Yes, it won't fix those use cases. Imo it is okay to just warn the user and leave it at that. 

Just curious, is there a case where one would want to make raw calls concurrently using a single connection? AFAIK db_connection is already lightweight from process footprint point of view.

Also, It's very hard to cover all such cases right? for example, parallel `sqlite3_close_v2` calls on the same connection. Currently we are not handling that I think even with mutex.

warmwaffles:
@greg-rychlewski are you suggesting that on the elixir side or the NIF side? 

On the NIF side, we'd have to use a mutex to ensure that a connection isn't double closed.

> Just curious, is there a case where one would want to make raw calls concurrently using a single connection? AFAIK db_connection is already lightweight from process footprint point of view.

I think someone who wants to skip using db_connection should be able to. But how realistic is it that someone would use the raw `Exqlite.Sqlite` interface outside of ecto? Probably low.

> Also, It's very hard to cover all such cases right? for example, parallel sqlite3_close_v2 calls on the same connection. Currently we are not handling that I think even with mutex.

Definitely hard to cover, but a mutex is probably the right solution to always check before calling `sqlite3_close_v2`

@akash-akya another thing is, could that mutex checking be adding enough latency to not surface the race condition? When I was throwing `fprintf` debug statements in, it slowed the calls down enough to make the issue not present itself.

akash-akya:
Yes, adding a delay in concurrent calls can make it hard to reach the exact race condition. I tried adding sleep before prepare and sure enough, it prevents the issue. 

The question now is if mutex is truly fixing the issue or just hiding the race condition because of latency it is adding right? I think if we can reliably prove the exact race condition sequence, then we can be sure about the fix.

Can you share your changes with `fprintf`? Since where we are placing it does matter

warmwaffles:
@akash-akya I shimmed it in as the first call in `exqlite_close` and one on `exqlite_open`.

akash-akya:
Trying to reproduce the issue deterministically for a single connection. But it is hard to get the order of thread execution correctly without any additional tooling.

Fortunately, since SQLite source code itself is embedded, we can just [add sleep in-between execution](https://github.com/elixir-sqlite/exqlite/commit/f01563fa1131b9de923c4a2fdbafc1edfd53eaf4#) of prepare & close, and get the deterministic order we want.

* branch with the deterministic test: https://github.com/akash-akya/exqlite/tree/deterministic-segfault
* branch with the deterministic test and fix: https://github.com/akash-akya/exqlite/tree/deterministic-segfault-with-fix

As we can see, the fix _does fixes_ the segfault in the test, and adding fprintf does not fix the above segfault.

Of course, among the set of all possible code path that can cause segfault, this is just one, there might be more.

akash-akya:
Yes, sounds good to me. 

Let me know if any more changes are pending for this to get merged.

warmwaffles:
Released under `v0.11.7`, thanks a ton @akash-akya

PR #219, opened 2022-10-30
Fix typos

kianmeng:
Found via `codespell -S c_src,deps`

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 An Elixir SQLite3 library.
 
-If you are looking for the Ecto adapater, take a look at the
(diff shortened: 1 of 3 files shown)


warmwaffles (approved):

warmwaffles:
I wonder if we should add that as a check to the lint action for github actions.

kianmeng:
> I wonder if we should add that as a check to the lint action for github actions.

Your call. :smile:

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

warmwaffles:
I'll explore it. I think it is worth having.

Issue #192, opened 2022-02-02
Replace DirtyNIF execution model with threadpool


warmwaffles:
The dirtynif execution model is okay for most cases, but we have an issue where long running writes need to timeout / be interrupted mid execution and aborted. The only way to do that is to utilize https://sqlite.org/c3ref/progress_handler.html

josevalim:
I wonder if you may be able to stick with the process model as long as you assume each instance of the database is backed by one process. This way you make Elixir processes your thread pool, instead of having to deal with it in C/C++/Rust.

warmwaffles:
@josevalim I think that's a good idea where only one process in the supervision tree sits there waiting for another `DbConnection` call. This _could_ make it easier to cancel existing running queries.

Let me noodle on this a little more.

josevalim:
Right, I think in this sense DBConnection may be getting more in the way than helping. Because you need serial access around SQLite3 and having a pool around serial access is not giving you anything. And when you don't need serial access (i.e. read operations), you don't need a pool at all.

DBConnection is necessary for Ecto but, even then, it should not be a requirement. There are only three features in Ecto.Adapters.SQL that require DBConnection (sandbox, stream, and disconnect_all) but they are all addressable.

Then there is the part of DBConnection as a convenience. I can think of two features it gives:

1. a transaction manager - if the process doing a transaction crashes, we need to abort it. but perhaps sqlite3 api itself already has conveniences for handling dangling transactions
2. the ownership pool

Of course there is a lot of work here, so I am not saying we should yank DBConnection, but it may be worth considering moving the DBConnection bits to ecto_sqlite3. In this project, based on the analysis above (which may be incomplete), you would only be missing the transaction manager (and it may be provided by sqlite3).

LostKobrakai:
There might be an additional benefit to managing transactions in a sqlite specific way. Currently there's no information returned to changesets about which constraint was violated, but there seems to be some pragmas to figure those out as sqlite doesn't close transactions on errors.

What I'm wondering about @josevalim's suggestion is how using processes to serialize or not serialize access would work for transactions which only become write transactions mid execution. Wouldn't the suggested approach require knowing if a transaction does writes or not in advance?

ruslandoga:
Wouldn't having one process per db instance negate WAL benefits? There still needs to be a pool of connections for concurrent reads. It's not possible to open a single connection and start multiple transactions on it.

josevalim:
[...]

PR #235, opened 2023-02-24
Add error logger

dangra:
I am submitting this change looking for directions on what could be the best way to integrate it into official exqlite release. This particular change has been extremely useful when debugging errors on applications that use [LiteFS](https://github.com/superfly/litefs) and I am sure could be useful to others.

From [sqlite docs](https://www.sqlite.org/c3ref/c_config_covering_index_scan.html):

> The SQLITE_CONFIG_LOG option is used to configure the SQLite global [error log](https://www.sqlite.org/errlog.html). (The SQLITE_CONFIG_LOG option takes two arguments: a pointer to a function with a call signature of void(*)(void*,int,const char*), and a pointer to void. If the function pointer is not NULL, it is invoked by [sqlite3_log()](https://www.sqlite.org/c3ref/log.html) to process each logging event. If the function pointer is NULL, the [sqlite3_log()](https://www.sqlite.org/c3ref/log.html) interface becomes a no-op.
>
>  The void pointer that is the second argument to SQLITE_CONFIG_LOG is passed through as the first parameter to the application-defined logger function whenever that function is invoked.
>
> The second parameter to the logger function is a copy of the first parameter to the corresponding [sqlite3_log()](https://www.sqlite.org/c3ref/log.html) call and is intended to be a [result code](https://www.sqlite.org/rescode.html) or an [extended result code](https://www.sqlite.org/rescode.html#extrc).
> 
> The third parameter passed to the logger is log message after formatting via [sqlite3_snprintf()](https://www.sqlite.org/c3ref/mprintf.html). 
>
> The SQLite logging interface is not reentrant; the logger function supplied by the application must not invoke any SQLite interface. In a multi-threaded application, the application-defined logger function must be threadsafe.


warmwaffles:
What we'll want to do here is not use that `fprintf`, but instead we'll need to take that error string that is emitted by sqlite, and notify erlang somehow with the message.

There is a function in the `sqlite3_nif.c` called `make_binary(env, the_string, length_of_string)`. This will return the `ERL_NIF_TERM` which we can then `enif_send` to a pid.

We'll need to stand up a listener for this.

This does tie into #192 somewhat where we should probably spin up one process instance.

warmwaffles:
@dangra What are you thinking about with consuming the log messages? Were you just wanting to dump to `Logger` and friends? Or were you wanting to monitor the events pushed out via a handler of some sort?

dangra:
@warmwaffles I haven't had the time to look at it and honestly lack the experience with NIFs and Elixir, to me it is all new stuff. Not making excuses! just setting expectations here, if the open PR bothers you, I am ok closing it for now and resubmit once I have something that integrates your feedback.

that said, my only goal was to log the messages so in case of this type of errors, I can tail logs and see what sqlite3.c function and line have error'ed. I don't have the need to push the events to a handler of any type.

warmwaffles:
@dangra I was going to implement this using the methods I outlined above. We shouldn't be printing to stdout using the NIF. We risk interleaving output in a highly concurrent environment.

ruslandoga:
👋

I've opened a PR with the suggested changes (send binary to pid): https://github.com/elixir-sqlite/exqlite/pull/266

I mostly wanted to make sure this feature would work after https://github.com/elixir-sqlite/exqlite/issues/192 and one way to do that is to put `add_error_log/2` in `master` :)

PR #240, opened 2023-02-27
docs: how to opt out of precompiled artefacts

cocoa-xu:
Hi @warmwaffles, I split the `Configuration` section into two subsections because I'd like to make it clearer that `make_force_build` is a compile-time option (so that users won't get confused about where they should set it). 

But please let me know if you have a better idea of which section I should mention this. :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,13 +42,26 @@ end
 
 ## Configuration
 
+### Runtime Configuration
+
 ```elixir
 config :exqlite, default_chunk_size: 100
 ```
 
 * `default_chunk_size` - The chunk size that is used when multi-stepping when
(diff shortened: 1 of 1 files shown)


warmwaffles (approved):

warmwaffles:
That's a good change. I wish there was a way to detect that a variable was modified at runtime that is compile time only and `Kernel.warn` developers.

cocoa-xu:
> That's a good change. I wish there was a way to detect that a variable was modified at runtime that is compile time only and `Kernel.warn` developers.

Yeah, I wish elixir has this feature as well.

PR #244, opened 2023-03-13
Added precompilation support for musl targets

cocoa-xu:
This PR updates `cc_precompiler` to `v0.1.7` and added three common musl targets,

- x86_64-linux-musl
- aarch64-linux-musl
- riscv64-linux-musl


warmwaffles:
You are awesome. Thank you @cocoa-xu!

PR #247, opened 2023-03-25
feat: enable load_extensions config for connections

bkono:
Opening this to start a discussion. This is an attempt to implement [this comment](https://github.com/elixir-sqlite/exqlite/issues/136#issuecomment-812802453) from issue #136, piggybacking on the hard work done in #160.

I needed this to be able to actively use extensions with Ecto inside a Phoenix app. There's a few scenarios where it comes in handy, but the primary driver for me is an app doing vector queries via [sqlite-vss](https://github.com/asg017/sqlite-vss/tree/main) which needs both vector0 and vss0 libs loaded on the connection.

With this change, adding the below in a phoenix app makes everything Just Work™:

```
config :exqlite,
  load_extensions: [
    "./priv/sqlite/darwin-arm64/lines0",
    "./priv/sqlite/darwin-arm64/vector0",
    "./priv/sqlite/darwin-arm64/vss0"
  ]
```

If there's interest in this, I'm happy to clean up the PR (*cough*IO.puts...*cough*) and update with documentation as well, but didn't want to invest too much time without confirming it's a change you want to take upstream.


bkono:
> This does introduce loading extensions globally for anyone using multiple ecto sqlite databases. While I think 90+% of use cases just have one DB, I think there will be others out there that need different extensions loaded for different repos.
> ....
> Having it at the global level for `exqlite` is good to have, but allowing it to be loaded on a per repo basis would be another configuration we could support.

Great suggestion. I've gone ahead and updated to support a combined global and per-connection options loading. Let me know if you have any codebase style preferences like extracting the case into separate do_load_extensions funs.

lib/exqlite/connection.ex
```diff
@@ -460,6 +461,30 @@ defmodule Exqlite.Connection do
...
+  defp load_extensions(db, options) do
+    global_extensions = Application.get_env(:exqlite, :load_extensions, [])
+
+    extensions =
+      Keyword.get(options, :load_extensions, [])
+      |> Enum.concat(global_extensions)
+      |> Enum.uniq()
+
+    case extensions do
+      [] ->
```
bkono, line 473: Figured it was worth the check and no-op, to avoid hitting the Sqlite3.enable_load_extension true/false sequence for an empty list.

lib/exqlite/connection.ex
```diff
@@ -480,7 +505,8 @@ defmodule Exqlite.Connection do
          :ok <- set_busy_timeout(db, options),
          :ok <- set_journal_size_limit(db, options),
          :ok <- set_soft_heap_limit(db, options),
-         :ok <- set_hard_heap_limit(db, options) do
+         :ok <- set_hard_heap_limit(db, options),
+         :ok <- load_extensions(db, options) do
```
bkono, line 509: Follows the fun(db, options) pattern now, looks much cleaner.

warmwaffles:
Perfect!

bkono:
> @bkono would you mind adding some documentation to the `Exqlite.Connection.connect/1` function? Once that's in place we should be ready to merge and release.

Of course.

image

In my apps I've been using a little helper for architecture resolution (since I'm on macOS & linux for dev & deploy). Wasn't sure if there's a good place to suggest something like this. I'll probably just create a light sample repo to point people at, but adding it here in case you have an alternative: 

```
arch_dir =
  System.cmd("uname", ["-sm"])
  |> elem(0)
  |> String.trim()
  |> String.replace(" ", "-")
  |> String.downcase()

config :myapp, arch_dir: arch_dir

config :myapp, Myapp.Repo,
  load_extensions: [
    "./priv/sqlite/#{arch_dir}/vector0",
    "./priv/sqlite/#{arch_dir}/vss0"
  ]
```

warmwaffles:
> In my apps I've been using a little helper for architecture resolution (since I'm on macOS & linux for dev & deploy). Wasn't sure if there's a good place to suggest something like this. I'll probably just create a light sample repo to point people at, but adding it here in case you have an alternative:

I think that is a totally reasonable example to provide. Others may do their resolutions differently.

bkono:
> > In my apps I've been using a little helper for architecture resolution (since I'm on macOS & linux for dev & deploy). Wasn't sure if there's a good place to suggest something like this. I'll probably just create a light sample repo to point people at, but adding it here in case you have an alternative:
> 
> I think that is a totally reasonable example to provide. Others may do their resolutions differently.

Done and updated. Let me know if there's anything else missing.

PR #251, opened 2023-06-04
check for closed connection in changes

mwhitworth:
Without this, it's possible to have a race condition where a crash happens if an insert query times out; the connection is closed during the `transaction_status`, `changes` follow-up

(Keen for feedback if this principle/check should be extended to any other NIF wrappers)

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -341,6 +341,10 @@ exqlite_changes(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         return make_error_tuple(env, "invalid_connection");
     }
 
+    if (conn->db == NULL) {
+        return make_error_tuple(env, "connection closed");
+    }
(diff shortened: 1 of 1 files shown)


c_src/sqlite3_nif.c
```diff
@@ -341,6 +341,10 @@ exqlite_changes(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         return make_error_tuple(env, "invalid_connection");
     }
 
+    if (conn->db == NULL) {
+        return make_error_tuple(env, "connection closed");
```
warmwaffles, line 345:
```suggestion
        return make_error_tuple(env, "connection_closed");
```

If you could also make this change to line 389 as well. That string gets turned into an atom. An oversight on my part.

warmwaffles (commented):
I've been needing to add this check in more places in the NIF code

PR #255, opened 2023-07-19
replace db_connection with rw lock

ruslandoga:
This PR begins to explore RW-lock approach suggested in https://github.com/elixir-sqlite/exqlite/issues/192#issuecomment-1634198546

Here's what I plan to try next:

- [ ] should `Exqlite.RWConnection.begin_transation(conn)` or `Exqlite.RWConnection.transaction(fun)` be supported?
- [ ] should `gen_statem` be used?
- [ ] should there be a cache for prepared statements?


josevalim:
>  should Exqlite.RWConnection.begin_transation(conn) or Exqlite.RWConnection.transaction(fun) be supported?

IMO, no transaction helpers necessary, unless it is required later by EctoSQLite3.

> should gen_statem be used?

Up to you :)

> should there be a cache for prepared statements?

I would postpone until there is a need, to keep the first PR small.

lib/exqlite/basic.ex
warmwaffles, line 1: This module _could_ be deleted and ignored until you are completely done. I don't know of anyone using this module directly.

warmwaffles:
>> should there be a cache for prepared statements?
>
> I would postpone until there is a need, to keep the first PR small.

I remember starting to cache the prepared statements, but then bailed on it due to the complexity of it. I figured since this is all in the NIF it was better to just create it on the fly and then let the garbage collector sweep it away once done.

ruslandoga:
Closing this PR as I think the processes can be moved to ecto_sqlite3 and exqlite would be just a NIF. I'm not 100% sure, but I'm exploring this approach in https://github.com/ruslandoga/sqlite.ex and https://github.com/ruslandoga/ecto_sqlite which a re basically copy-pastes of exqlite/ecto_sqlite3 with some changes here and there (and different project / namespace makes benchmarks easier).

PR #256, opened 2023-07-21
rw conns

ruslandoga:
This PR is similar to #255 but instead of making writes exclusive it opens two databases, one for writing and one for reading. This allows for reads to continue being served while the write is happening thanks to [WAL:](https://www.sqlite.org/wal.html)

> 2. WAL provides more concurrency as readers do not block writers and a writer does not block readers. Reading and writing can proceed concurrently.


lib/exqlite.ex
```diff
@@ -3,102 +3,24 @@ defmodule Exqlite do
...
-  alias Exqlite.Error
-  alias Exqlite.Query
+  alias Exqlite.RWConnection
   alias Exqlite.Result
 
-  @doc "See `Exqlite.Connection.connect/1`"
-  @spec start_link([Connection.connection_opt()]) :: {:ok, pid()} | {:error, Error.t()}
-  def start_link(opts) do
-    DBConnection.start_link(Connection, opts)
-  end
```
ruslandoga, line 15: Most of the public API would need to be restored before merging if this experiment is successful.

warmwaffles:
Some issues that come to mind that will need to be explored.

1. Native extension support. Can it be done with the two database approach?
2. Will this work with `sqlcipher` compiled with the library? We'll need to be able to pass configurations for it.
3. Will we be able to support all of the original connection options prior to this?

lib/exqlite/rw_connection.ex
```diff
@@ -0,0 +1,274 @@
...
+         do: {:ok, %Exqlite.Result{columns: columns, rows: rows}}
+  end
+
+  @impl true
+  def init(options) do
+    database = Keyword.fetch!(options, :database)
+
+    if database in [:memory, ":memory:"] do
+      raise "#{database} is not supported in #{__MODULE__}"
+    end
```
warmwaffles, line 98: I think it could be supported. SQLite has internal locking mechanisms that should handle two concurrent reads and writes in memory no problem. 

https://www.sqlite.org/c3ref/threadsafe.html

We have it compiled with

```
CFLAGS += -DSQLITE_THREADSAFE=1
```

ruslandoga:
@warmwaffles 

1. Yeah... `select load_extension(..., ...)` would need to be called on both dbs, not great.

```
def load_extension(conn, extension) dp
  with {:ok, _} <- query(conn, "select load_extension(#{extension})"),
       {:ok, _} <- read_query(conn, "select load_extension(#{extension})"), do: :ok
end
```

2. I think this is similar to 1 and

lib/exqlite/rw_connection.ex
ruslandoga: I thought the databases would be in different places in memory. So inserting in one wouldn't make the data visible in the other.

warmwaffles: They will be in different places yes. I forgot to send along that I think we could support an in memory database by setting both the reader and writer db instances to be the same DB handles.

ruslandoga: Ah, right. I'll try that.

warmwaffles: I think you'd just need to track if it was opened as in memory so you know to close the db once and then dereference the old handle.

ruslandoga:
Closing this PR as it's too complicated.

warmwaffles:
Hah, it's not a problem. I'll find some time to really dive into this more.

PR #258, opened 2023-08-17
fix: don't bind atoms that are prefixed with `nil` as NULL

mhanberg:
This patch changes the comparison from `utf8ncmp` to `utf8cmp` in order
to explicitly match on the size of the atom. I noticed a case of
inserting an atom of `:nil_safe?` that was cast to `NULL`.

I wasn't sure the correct level to add a test, so please let me know the
best place and I can get it added.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -430,7 +430,7 @@ bind(ErlNifEnv* env, const ERL_NIF_TERM arg, sqlite3_stmt* statement, int index)
     }
 
     if (enif_get_atom(env, arg, the_atom, sizeof(the_atom), ERL_NIF_LATIN1)) {
-        if (0 == utf8ncmp("undefined", the_atom, 9) || 0 == utf8ncmp("nil", the_atom, 3)) {
+        if (0 == utf8cmp("undefined", the_atom) || 0 == utf8cmp("nil", the_atom)) {
             return sqlite3_bind_null(statement, index);
         }
 


warmwaffles:
Oh what an interesting problem. I wanted to cap the check of two raw strings, I completely wiffed on this and should have realized that of course `:nil_safe?` compared to `:nil` with 3 would always equal 0.

ruslandoga:
I think it might make sense to get rid of utf8 functions completely as both SQLite and nifs expect just the count of bytes, not utf8 codepoints. For example, getting columns for a table like

```sql
sqlite> create table 👋(✍️ text,\U+1FAF1 integer) strict;
Parse error: unknown datatype for 👋.✍️: "text🫱 integer"
sqlite> create table 👋(✍️ text, \U+1FAF1 integer) strict;
sqlite> select * from sqlite_master;
table|👋|👋|2|CREATE TABLE 👋(✍️ text, 🫱 integer) strict
sqlite> .mode box
sqlite> insert into 👋 values ('\U+1FAF1', 10);
sqlite> select * from 👋;
┌────┬────┐
│ ✍️ │ 🫱  │
├────┼────┤
│ 🫱  │ 10 │
└────┴────┘
```

results in invalid data returned in `exqlite`

```elixir
iex(1)> {:ok, db} = Exqlite.Sqlite3.open("demo.db")
{:ok, #Reference<0.3292708811.1647181846.21903>}
iex(2)> Exqlite.Sqlite3.prepare(db, "select * from 👋")
{:ok, #Reference<0.3292708811.1647181846.21905>}
iex(3)> {:ok, stmt} = v
{:ok, #Reference<0.3292708811.1647181846.21905>}
iex(4)> Exqlite.Sqlite3.columns(db, stmt)
{:ok, [<<226, 156>>, <<240>>]}
iex(5)> "✍️"
"✍️"
iex(6)> "✍️" <> <<0>>
<<226, 156, 141, 239, 184, 143, 0>>
```

Issue #221, opened 2022-11-08
Feature request: Add support for Data Change Notification Callbacks


OldhamMade:
It would be great if exqlite supported Data Change Notification Callbacks, eg. `set_update_hook`, similar to [mmzeeman/esqlite](https://github.com/mmzeeman/esqlite). `esqlite` is great, however it is very Erlang-focused and the original Elixir wrapper library for it has been abandoned. It would be very useful to be able to act on hooks within a project that is using [ecto_sqlite3](https://hex.pm/packages/ecto_sqlite3).

ruslandoga:
@warmwaffles I wonder if it can be a separate connection behaviour like https://github.com/elixir-ecto/postgrex/blob/master/lib/postgrex/replication_connection.ex.

warmwaffles:
I'm not entirely sure how this would function because the update hooks would only work on the currently connected db connection / exqlite connection.

Definitely an interesting use case that doesn't fit the current paradigm that `DbConnection` and `Ecto` have defined.

OldhamMade:
> Can you provide me a concrete use case that this would fit in to?

It would allow someone using a SQLite database to get close to mimicking the LISTEN/NOTIFY functionality from Postgres, which is my specific use-case: I'd like to act on table updates caused by DB triggers from within Elixir code.

> Definitely an interesting use case that doesn't fit the current paradigm that DbConnection and Ecto have defined.

I'm not sure that it needs to fit with that paradigm. I think it is more akin to `Postgrex.Notifications`, a useful extra.

warmwaffles:
I didn't realize `Postgrex` had that ability. I'll take a look at how they resolved the `LISTEN` / `NOTIFY`. We may be able to utilize a similar functionality.

mdwaud:
+1 for interest in this capability

PR #260, opened 2023-09-24
Add set_update_hook/2

greg-rychlewski:
Closes https://github.com/elixir-sqlite/exqlite/issues/221

I took a stab at this based on how the [erlang sqlite library is handling it](https://github.com/mmzeeman/esqlite/).

The implementation is pretty straight forward. The user supplies a connection and a pid and we create a callback that sends a message to the pid of the form `{:insert | :update | :delete, db_name, table, row_id}` (that's all the information SQLite exposes about the change. For example, you could create a GenServer to listen to the notifications.

Right now, there can only be one pid per database connection. But I think it can probably be extended later to work with a list of pids.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -1061,7 +1061,7 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 
     // Passing the connection as the third argument causes it to be
     // passed as the first argument to update_callback. This allows us
-    // To extract the hook pid and reset the hook if the pid is not alive.
+    // to extract the hook pid and reset the hook if the pid is not alive.
     sqlite3_update_hook(conn->db, update_callback, conn);
 
     return make_atom(env, "ok");
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -16,13 +16,14 @@
 #define MAX_PATHNAME    512
 
 static ErlNifResourceType* connection_type = NULL;
-static ErlNifResourceType* statement_type  = NULL;
+static ErlNifResourceType* statement_type = NULL;
 static sqlite3_mem_methods default_alloc_methods = {0};
 
 typedef struct connection
 {
     sqlite3* db;
     ErlNifMutex* mutex;
+    ErlNifPid update_hook_pid;
 } connection_t;
 
 typedef struct statement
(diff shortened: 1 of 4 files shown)


greg-rychlewski:
This is exposed on the driver level but I think if someone really wanted to use it with Ecto they could use the `after_connect` configuration option and supply `{Exqlite.Sqlite3, :set_update_hook, [genserver_name]}`.

I believe this will work and also be called on reconnects but haven't tested. That's why I didn't advertise it in the docs yet.

lib/exqlite/sqlite3.ex
```diff
@@ -203,6 +203,32 @@ defmodule Exqlite.Sqlite3 do
     end
   end
 
+  @doc """
+  Send data change notifications on connection to a process.
```
greg-rychlewski, line 207:
```suggestion
  Send data change notifications to a process.
```

c_src/sqlite3_nif.c
```diff
@@ -16,13 +16,14 @@
...
-static ErlNifResourceType* statement_type  = NULL;
+static ErlNifResourceType* statement_type = NULL;
 static sqlite3_mem_methods default_alloc_methods = {0};
 
 typedef struct connection
 {
     sqlite3* db;
     ErlNifMutex* mutex;
+    ErlNifPid update_hook_pid;
 } connection_t;
```
warmwaffles, line 27: I wonder if we could capture the `App.Repo` atom and store it with the connection and pass that in the callback instead of the raw db name. This would probably change how the sqlite database is opened.

lib/exqlite/sqlite3.ex
```diff
@@ -203,6 +203,32 @@ defmodule Exqlite.Sqlite3 do
...
+    * `row_id` is an integer representing the unique row id assigned by SQLite
+
+  ## Restrictions
+
+    * There are some conditions where the update hook will not be invoked by SQLite.
+      See the documentation for [more details](https://www.sqlite.org/c3ref/update_hook.html)
+    * Only one pid can listen to the changes on a given database connection at a time.
+      If this function is called multiple times for the same connection, only the last pid will
+      receive the notifications
+  """
```
warmwaffles, line 226: I think we will probably need to add something to the effect of

```suggestion
    * Updates only happen for the connection that is opened. For example, there 
      are two connections A and B. When an update happens on connection B, the
      hook set for connection A will not receive the update, but the hook for
      connection B will receive the update.
  """
```

warmwaffles (approved):
This is good. I think this is probably the simplest way to implement this without having to do some changes on connection construction.

greg-rychlewski:
Oh I was looking into whether Ecto passes the repo into the `connect` function. Would you be interested in a separate PR for that if it works?

warmwaffles:
Yea I think that would be fine. I have a feeling it may be a breaking api change. Not entirely sure.

sbaildon:
Actually, this works just fine but I feel like I'm not meant to be doing this:


```elixir
  def after_connect(%{pool_ref: {:pool_ref, pool, _, nil, _, nil}}, receiver_pid) when is_pid(pool)  do
    with {_, %{state: %{db: ref}}} when is_reference(ref) <- :sys.get_state(pool) do
        :ok = Exqlite.Sqlite3.set_update_hook(ref, receiver_pid)
      end
  end
```

PR #263, opened 2023-10-11
use size_t to store allocated size

### LINKED ISSUE #209: [question] sqlite extension alignment problem in exqlite, but not in sqlite3
👋

I understand that it's unlikely to be a problem in `exqlite`, but just in case you have any idea as to what I'm doing wrong.

---

Repo: https://github.com/ruslandoga/sqlite-zig-problem

I'm trying to write an extension for sqlite in [zig](https://ziglang.org) and I'm getting an [alignment panic](https://ziglang.org/documentation/master/#Alignment) when fetching sqlite's aggregate context. But, I'm only getting it when executing the query from elixir, not from `sqlite3` cli.

Meaning,

```sql
sqlite> .load dist/extc
sqlite> .load dist/extzig

-- this uses c extension
sqlite> with recursive ten(x) as (select 1 union all select x+1 from ten where x<10) select sumc(x) from ten;
55.0

-- this uses zig extension
sqlite> with recursive ten(x) as (select 1 union all select x+1 from ten where x<10) select sumzig(x) from ten;
55.0
```

but in elixir:

```elixir
iex(1)> E.sumc
# [debug] QUERY OK db=0.6ms decode=1.3ms idle=738.5ms
# with recursive ten(x) as (select 1 union all select x+1 from ten where x<10) select sumc(x) from ten []
%Exqlite.Result{
  columns: ["sumc(x)"],
  command: :execute,
  num_rows: 1,
  rows: [[55.0]]
}

iex(2)> E.sumzig
thread 8760720 panic: incorrect alignment
```

ruslandoga:
fixes #209

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -37,7 +37,7 @@ exqlite_malloc(int bytes)
 {
     assert(bytes > 0);
 
-    int* p = enif_alloc(bytes + sizeof(int));
+    size_t* p = enif_alloc(bytes + sizeof(size_t));
     if (p) {
         p[0] = bytes;
         p++;
@@ -53,7 +53,7 @@ exqlite_free(void* prior)
         return;
     }
 
-    int* p = prior;
+    size_t* p = prior;
 
     // Shift the pointer back to free the proper block of data
     p--;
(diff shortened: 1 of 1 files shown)


warmwaffles:
I wonder if there is a test we can do to ensure this doesn't happen again. I _believe_ this is the right solution. I _think_ the only way would be to compile in arm64 and run the tests on that architecture. Wonder how hard that is to do with github actions?

ruslandoga:
I might be wrong but I don't think it's arm64 specific. The pointer was misaligned on x86_64 as wel, it's just that C doesn't care about it and Zig did (I guess because they  use SSE instructions in their standard lib)

warmwaffles:
Ah

PR #264, opened 2023-10-12
drop utf8.h

ruslandoga:
as proposed in https://github.com/elixir-sqlite/exqlite/pull/258#issuecomment-1758983541


Makefile
```diff
@@ -19,13 +19,11 @@
 #
 
 SRC = c_src/sqlite3_nif.c
-HEADERS = c_src/utf8.h
 
 CFLAGS = -I"$(ERTS_INCLUDE_DIR)"
 
 ifeq ($(EXQLITE_USE_SYSTEM),)
 	SRC += c_src/sqlite3.c
-	HEADERS += c_src/sqlite3.h c_src/sqlite3ext.h
```
ruslandoga, line 28: It seems like `HEADERS` were unused.

warmwaffles (approved):
Thank you for adding tests as well.

PR #266, opened 2023-10-19
add set_log_hook/1

ruslandoga:
Continues from https://github.com/elixir-sqlite/exqlite/pull/235

Based on #260

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -1066,6 +1066,56 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 }
 
 //
+// Log Notifications
+//
+
+ErlNifPid* log_hook_pid = NULL;
+
+void
+log_callback(void* arg, int iErrCode, const char* zMsg)
+{
+    if (log_hook_pid == NULL) {
+        return;
+    }
+
+    ErlNifEnv* msg_env = enif_alloc_env();
+    ERL_NIF_TERM error = make_binary(msg_env, zMsg, strlen(zMsg));
+    ERL_NIF_TERM msg   = enif_make_tuple3(msg_env, make_atom(msg_env, "log"), enif_make_int(msg_env, iErrCode), error);
+
+    if (!enif_send(NULL, log_hook_pid, msg_env, msg)) {
+        sqlite3_config(SQLITE_CONFIG_LOG, NULL, NULL);
+        log_hook_pid = NULL;
+        enif_free(log_hook_pid);
+    }
+
+    enif_free_env(msg_env);
+}
+
+static ERL_NIF_TERM
+exqlite_set_log_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
+{
+    assert(env);
+
+    if (argc != 1) {
+        return enif_make_badarg(env);
+    }
+
+    ErlNifPid* pid = (ErlNifPid*)enif_alloc(sizeof(ErlNifPid));
+    if (!enif_get_local_pid(env, argv[0], pid)) {
+        enif_free(pid);
+        return make_error_tuple(env, "invalid_pid");
+    }
+
+    if (log_hook_pid) {
+        enif_free(log_hook_pid);
+    }
+
+    log_hook_pid = pid;
+    sqlite3_config(SQLITE_CONFIG_LOG, log_callback, NULL);
+    return make_atom(env, "ok");
+}
+
+//
 // Most of our nif functions are going to be IO bounded
 //
 
@@ -1086,6 +1136,7 @@ static ErlNifFunc nif_funcs[] = {
   {"release", 2, exqlite_release, ERL_NIF_DIRTY_JOB_IO_BOUND},
   {"enable_load_extension", 2, exqlite_enable_load_extension, ERL_NIF_DIRTY_JOB_IO_BOUND},
   {"set_update_hook", 2, exqlite_set_update_hook, ERL_NIF_DIRTY_JOB_IO_BOUND},
+  {"set_log_hook", 1, exqlite_set_log_hook, ERL_NIF_DIRTY_JOB_IO_BOUND},
 };
 
 ERL_NIF_INIT(Elixir.Exqlite.Sqlite3NIF, nif_funcs, on_load, NULL, NULL, on_unload)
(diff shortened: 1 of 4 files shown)


test/exqlite/sqlite3_test.exs
```diff
@@ -491,4 +491,68 @@ defmodule Exqlite.Sqlite3Test do
...
+        assert msg == "near \"some\": syntax error in \"some invalid sql\""
+        refute_receive _anything_else
+        send(test, :done)
+      end)
+
+      assert_receive :done, 1000
+      refute_receive _anything_else
+    end
+
+    test "notifications don't cross connections", %{conn: conn1, path: path} do
```
ruslandoga, line 539: I wonder how this works though, since the `sqlite_config` call seems to be global as it doesn't require a DB reference in its args.

c_src/sqlite3_nif.c
```diff
@@ -1065,6 +1066,55 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    }
+
+    if (!enif_get_local_pid(env, argv[1], &conn->log_hook_pid)) {
+        return make_error_tuple(env, "invalid_pid");
+    }
+
+    // Passing the connection as the third argument causes it to be
+    // passed as the first argument to log_callback. This allows us
+    // to extract the hook pid and reset the hook if the pid is not alive.
+    sqlite3_config(SQLITE_CONFIG_LOG, log_callback, conn);
```
ruslandoga, line 1114: Here, notice no `conn->db` passed in.

c_src/sqlite3_nif.c
```diff
@@ -1065,6 +1065,56 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     return make_atom(env, "ok");
 }
 
+//
+// Log Notifications
+//
+
+ErlNifPid* log_hook_pid = NULL;
```
ruslandoga, line 1072: There can only be one, hence a global variable. Maybe we can add a mutex too.

ruslandoga: Regarding the mutex, I think the use-case would be to set it on app startup once, so the mutex there is not necessary.

```elixir
defmodule Some.Application do
  use Application
  def start(_, _) do
     :ok  = Exqlite.Sqlite3.set_log_hook(some_logger_pid)
     # etc.
   end
end
```

warmwaffles: Definitely need a mutex here. https://www.erlang.org/doc/man/erl_nif#enif_mutex_create

warmwaffles: Also can you lift the global to be near the top of the file?

ruslandoga: done in https://github.com/elixir-sqlite/exqlite/pull/266/commits/919d2ef91fc3ad78ad583ab68d0c11686fe3239d

c_src/sqlite3_nif.c
```diff
@@ -1065,6 +1068,69 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    ErlNifEnv* msg_env = enif_alloc_env();
+    ERL_NIF_TERM error = make_binary(msg_env, zMsg, strlen(zMsg));
+    ERL_NIF_TERM msg   = enif_make_tuple3(msg_env, make_atom(msg_env, "log"), enif_make_int(msg_env, iErrCode), error);
+
+    if (!enif_send(NULL, log_hook_pid, msg_env, msg)) {
+        enif_mutex_lock(log_hook_mutex);
+        sqlite3_config(SQLITE_CONFIG_LOG, NULL, NULL);
+        enif_free(log_hook_pid);
+        log_hook_pid = NULL;
+        enif_mutex_unlock(log_hook_mutex);
```
ruslandoga, line 1091: I'm still using the mutex when resetting the config so no `enif_mutex_destroy`

c_src/sqlite3_nif.c
```diff
@@ -1065,6 +1068,69 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    }
+
+    enif_mutex_lock(log_hook_mutex);
+
+    if (log_hook_pid) {
+        enif_free(log_hook_pid);
+    }
+
+    log_hook_pid = pid;
+    sqlite3_config(SQLITE_CONFIG_LOG, log_callback, NULL);
```
warmwaffles, line 1127: It's a shame this log hook is global like this. I wish it were on a per sqlite database basis, but alas it is not. But I supposed you would only utilize this in a dev or test environment to track down a bug.

ruslandoga: The reason it's not per database is I guess to also be able to cover errors coming from opening invalid databases. The docs mention that it's better to use this in production, see https://www.sqlite.org/errlog.html

c_src/sqlite3_nif.c
```diff
@@ -1065,6 +1068,69 @@ exqlite_set_update_hook(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+        return make_error_tuple(env, "invalid_pid");
+    }
+
+    if (log_hook_mutex == NULL) {
+        log_hook_mutex = enif_mutex_create("exqlite:log_hook");
+
+        if (log_hook_mutex == NULL) {
+            return make_error_tuple(env, "failed_to_create_mutex");
+        }
+    }
```
warmwaffles, line 1118: I believe this still has a chance of a race condition. This should be chucked into the `on_load` function and then the teardown of the mutex thrown into `on_unload`.

ruslandoga: Do you mean multiple threads calling `enif_mutex_create`? It's marked as thread safe in the docs.

warmwaffles: Outside of this, I am curious what the behavior is when two threads call `enif_mutex_create` with the same parameter. Obviously the one that called it first will win, but what happens to the loser?

ruslandoga: Ah, it probably should've been

```c
ErlNifMutex* mutex = enif_mutex_create("exqlite:log_hook");

if (mutex) {
  log_hook_mutes = mutex;
}  else {
  return make_error_tuple(env, "failed_to_create_mutex");
}
```

ruslandoga: Moved mutex to on_load in https://github.com/elixir-sqlite/exqlite/pull/266/commits/dc138b2744f8ad9fbbdf488ba3a5b7f39338926c

PR #270, opened 2023-10-30
Update README.md

dvic:
I'm not sure if the usage of BLOB is documented elsewhere but I've added it to the readme.


warmwaffles:
Thank you. If you see any more documentation that needs updating or clarification, feel free to open a PR or an issue and we can get that resolved.

PR #271, opened 2023-11-02
Adding before_disconnect hook

jeregrine:
This is useful with some extensions like [cr-sqlite](https://github.com/vlcn-io/cr-sqlite) that require a cleanup setup before disconnect/close, which is not [un-common](https://sqlite.org/forum/forumpost/c94f943821). 

I thought maybe this would work in db_connection but I don't know if its general enough for ALL db connections and ahh touching that state machine is the worst.

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -36,7 +36,8 @@ defmodule Exqlite.Connection do
     :path,
     :transaction_status,
     :status,
-    :chunk_size
+    :chunk_size,
+    :before_disconnect
   ]
 
   @type t() :: %__MODULE__{
@@ -44,7 +45,9 @@ defmodule Exqlite.Connection do
           directory: String.t() | nil,
           path: String.t(),
           transaction_status: :idle | :transaction,
-          status: :idle | :busy
+          status: :idle | :busy,
+          chunk_size: integer(),
+          before_disconnect: (t -> any) | {module, atom, [any]} | nil
         }
 
   @type journal_mode() :: :delete | :truncate | :persist | :memory | :wal | :off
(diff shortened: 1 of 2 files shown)


test/exqlite/connection_test.exs
```diff
@@ -184,6 +184,19 @@ defmodule Exqlite.ConnectionTest do
...
+      {:ok, conn} =
+        Connection.connect(
+          database: :memory,
+          before_disconnect: fn err, db ->
+            assert err
+            assert is_function(db.before_disconnect)
+          end
+        )
+
+      assert :ok == Connection.disconnect(true, conn)
```
warmwaffles, line 198: I _think_ what we want to do here as well is spin up a temporary `Agent` initialize it to 0, then inside the before disconnect, bump it +1, then assert that the value in the agent is 1 after disconnect is called.

I don't know if the concept of test spys are available in ex unit.

lib/exqlite/connection.ex
```diff
@@ -155,6 +159,7 @@ defmodule Exqlite.Connection do
             "./priv/sqlite/\#{arch_dir}/vss0"
           ]
       ```
+    * `:before_disconnect` - A function to run before disconnect, either a 2-arity fun or `{module, function, args}` with the close reason and `t:Exqlite.Connection.t/0` prepended to `args` or `nil` (default: `nil`)
```
warmwaffles, line 162: Need to wrap the docs.

```suggestion
    * `:before_disconnect` - A function to run before disconnect, either a 
      2-arity fun or `{module, function, args}` with the close reason and 
      `t:Exqlite.Connection.t/0` prepended to `args` or `nil` (default: `nil`)
```

jeregrine:
<3 for prompt feedback

warmwaffles (approved):

warmwaffles:
@jeregrine I'll be cutting a release for this soon.

warmwaffles:
This will be released under `v0.17.0`

PR #273, opened 2023-12-22
drop db_connection

ruslandoga:
Since there is [interest](https://github.com/elixir-sqlite/exqlite/issues/192#issuecomment-1866504011) in moving `db_connection` to `ecto_sqlite3` I thought I'd PR this change :)

TODOs:
- finish tests
- finish TODOs in code
- add individual binds (done in https://github.com/ruslandoga/xqlite)
- add prepare_v3 (done in https://github.com/ruslandoga/xqlite)
- explain the NIF changes
- move benchmarks elsewhere


README.md
```diff
@@ -15,42 +15,25 @@ Package: https://hex.pm/packages/exqlite
...
-  it will be interpreted as a string. 
+  it will be interpreted as a string.
 
 ## Installation
 
 ```elixir
 defp deps do
   [
+    # TODO
     {:exqlite, "~> 0.17"}
```
ruslandoga, line 31: - [ ] Version would need to be bumped.

README.md
```diff
@@ -15,42 +15,25 @@ Package: https://hex.pm/packages/exqlite
...
 ## Caveats
 
-* Prepared statements are not cached.
 * Prepared statements are not immutable. You must be careful when manipulating
   statements and binding values to statements. Do not try to manipulate the
   statements concurrently. Keep it isolated to one process.
-* Simultaneous writing is not supported by SQLite3 and will not be supported
-  here.
-* All native calls are run through the Dirty NIF scheduler.
-* Datetimes are stored without offsets. This is due to how SQLite3 handles date
```
ruslandoga, line 25: I think the Elixir -> SQLite3 mapping seems more like ecto_sqlite3 concern. `Exqlite` can focus on SQLite3 data types.

README.md
```diff
@@ -15,42 +15,25 @@ Package: https://hex.pm/packages/exqlite
...
 end
 ```
 
-
 ## Configuration
-
-### Runtime Configuration
-
-```elixir
-config :exqlite, default_chunk_size: 100
```
ruslandoga, line 49: This can be configured per repo in ecto_sqlite3.

README.md
```diff
@@ -115,38 +98,30 @@ export EXQLITE_SYSTEM_CFLAGS=-I/usr/local/include/sqlcipher
 export EXQLITE_SYSTEM_LDFLAGS=-L/usr/local/lib -lsqlcipher
 ```
 
-Once you have `exqlite` configured, you can use the `:key` option in the database config to enable encryption:
+Once you have `exqlite` build configured, you can use the `key` pragma to enable encryption:
 
 ```elixir
-config :exqlite, key: "super-secret'
+{:ok, conn} = Exqlite.open("sqlcipher.db")
+:ok = Exqlite.execute(conn, "pragma key='super-secret'")
```
ruslandoga, line 105: I've removed pragma helpers for a simpler API. Personally, I don't see much value in hiding the pragma statements from the user.

c_src/sqlite3_nif.c
```diff
@@ -859,17 +859,12 @@ exqlite_release(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     assert(env);
 
     statement_t* statement = NULL;
-    connection_t* conn     = NULL;
```
ruslandoga, line 862: `sqlite3_finalize` doesn't need `conn`

README.md
ruslandoga: Also I'm getting this error when compiling

```
==> exqlite
Error happened while installing exqlite from precompiled binary: "missing checksum.exs file".

Attempting to compile exqlite from source...
 CC sqlite3_nif.o
 CC sqlite3.o
 LD sqlite3_nif.so
```

Not sure if it's caused by my changes. I'll investigate.

warmwaffles: For this major move I'd bump major to `1.0` since it is a breaking change.

lib/exqlite/nif.ex
```diff
@@ -0,0 +1,30 @@
...
+  def multi_bind_step(_conn, _stmt, _rows), do: :erlang.nif_error(:not_loaded)
+  def columns(_conn, _stmt), do: :erlang.nif_error(:not_loaded)
+  def last_insert_rowid(_conn), do: :erlang.nif_error(:not_loaded)
+  def transaction_status(_conn), do: :erlang.nif_error(:not_loaded)
+  def serialize(_conn, _database), do: :erlang.nif_error(:not_loaded)
+  def deserialize(_conn, _database, _serialized), do: :erlang.nif_error(:not_loaded)
+  def release(_stmt), do: :erlang.nif_error(:not_loaded)
+  def enable_load_extension(_conn, _flag), do: :erlang.nif_error(:not_loaded)
+  def set_update_hook(_conn, _pid), do: :erlang.nif_error(:not_loaded)
+  def set_log_hook(_pid), do: :erlang.nif_error(:not_loaded)
```
warmwaffles, line 29: I would like to retain the typespecs for these functions.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unlreleased
 
+- removed: `db_connection`. It's just a NIF now.
```
warmwaffles, line 5: Also needs a note about removing `Exqlite.Basic`

lib/exqlite.ex
```diff
@@ -3,102 +3,274 @@ defmodule Exqlite do
...
+    nofollow: 0x01000000,
+    exrescode: 0x02000000
+  ]
 
-      otherwise ->
-        otherwise
-    end
+  open_flag_names = Enum.map(open_flag_mappings, fn {name, _value} -> name end)
+  open_flag_union = Enum.reduce(open_flag_names, &{:|, [], [&1, &2]})
+  @type open_flag :: unquote(open_flag_union)
```
warmwaffles, line 41: Oh this is cool, I didn't realize you could do this with typespecs.

lib/exqlite/nif.ex
ruslandoga: Sure, but may I ask why? This module is supposed to be hidden, and the public API is already types. Also, it seems like the previous types were incorrect, e.g. String.Chars.t instead of iodata and it wasn't being caught.

warmwaffles: You are right. The module is still public since there is no notion of `protected` in elixir / erlang. `@moduledoc false` won't have it show up, but I don't know if the dialyzer will honor that.

README.md
warmwaffles: My original intent behind the pragma helpers was to provide a more elixir / erlang interface to manipulate sqlite. Now with this setup, the caller will have to construct the strings to execute manually, which may or may not be a bad thing. I also wanted to be able to allow developers to configure exqlite via `config/*.exs` rather than having to make a factory module to produce and set configurations before utilizing the connection.

Now with removing the configurations, you are responsible for configuring your connection. Which is fine if `ecto_sqlite3` allows the pragmas to be set via configs.

```elixir
defmodule MyApp.ExqliteFactory do
  def build_conn do
    configs = Application.get_env(:my_app, Exqlite)

    {:ok, conn} = Exqlite.open(configs[:db])

    if configs[:key] do
      :ok = Exqlite.execute(conn, "pragma key='super-secret'")
    end

    # more connetion configuration setup here

    {:ok, conn}
  end
end
```

ruslandoga: With Ecto it would be something like

```config
config :my_app, MyApp.Repo,
  after_connect: fn conn ->
    :ok = Exqlite.execute(conn, "pragma key='super-secret'")
  end
```

warmwaffles: Makes sense to me. We'll need to document that as a solution in the README for those who want to alter the pragmas.

warmwaffles:
> https://github.com/elixir-sqlite/exqlite/pull/271 functionality is "lost", what should be done about it?

Would it be useful to perhaps lift the callback definition to the same place `after_connect` is specified? https://github.com/elixir-ecto/db_connection/blob/fa5f705fa5d272ed28b64ee0954e4275c0260d36/lib/db_connection.ex#L118-L135 SQLite users can't be the only ones who would find a callback to fire before disconnect as useful.

josevalim:
> SQLite users can't be the only ones who would find a callback to fire before disconnect as useful.

My concern about such callbacks is that we can't ever rely on them running though (what if the whole node terminates), so I worry people may rely on it always executing. It probably makes more sense to setup a monitor process instead or similar.

warmwaffles:
@ruslandoga It's not a problem. I too, have had other priorities take over where I can't find time to carve out and think hard about this.

c_src/sqlite3_nif.c
```diff
@@ -1157,29 +1221,36 @@ exqlite_interrupt(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+  {"execute", 2, exqlite_execute},
+  {"changes", 1, exqlite_changes},
+  {"prepare", 2, exqlite_prepare},
+  {"columns", 2, exqlite_columns},
+  {"step", 2, exqlite_step},
+  {"interrupt", 1, exqlite_interrupt},
+  {"finalize", 1, exqlite_finalize},
+  {"last_insert_rowid", 1, exqlite_last_insert_rowid},
+  {"transaction_status", 1, exqlite_transaction_status},
```
warmwaffles, line 1247: Why are these not going through the Dirty NIF? These are io bound, specifically `exqlite_step` is definitely IO bound.

ruslandoga: 👋 @warmwaffles 

Not necessarily. Sometimes `exqlite_step` can be safely scheduled on non-dirty thread, this change introduces multiple versions of each nif, making it possible for the caller to decide which one to use. By default, the library would keep using dirty schedulers but it will also introduce a family of "unsafe" implementations :)

ruslandoga: I decided to do this split when I noticed that non-dirty exqlite_step is almost twice as performant as a dirty exqlite_multi_step on all data sets I tried them on.

ruslandoga: > I'll look into simulating slow disk access (system under heavy write mode)

And I'll try to find their implementation. Right now I think they just read from internal connection state.

warmwaffles: > I decided to do this split when I noticed that non-dirty `exqlite_step` is almost twice as performant as a dirty `exqlite_multi_step` on all data sets I tried them on.

I originally did the multi step implementation because I had a very large sqlite database of timeseries data and it did not have a lot of throughput. But of course that was on the implementation I wrote, and what you have here _may_ actually be better.

warmwaffles:
@ruslandoga do you have a branch on `ecto_sqlite3` for changes made here? I'd like to give it a whirl if you do.

ruslandoga:
No, sorry. I haven't touched `ecto_sqlite3` yet. I'll try to come up with something tomorrow :)

warmwaffles:
> I'll try to come up with something tomorrow

No rush, I can work with the raw NIF until then.

warmwaffles:
One interesting thing to consider about moving the DB Connection to `ecto_sqlite3`. I would love to be able to run a repo with `:memory:` and it stay alive for as long as the application is up. My use case here is a queryable cache.

warmwaffles:
Sorry @ruslandoga I changed a bit of the code in around the NIF and sqlite to raise an exception when binding arguments to a statement. I wonder if we can accomplish this move to remove db connection in parts rather than one big change.

Mainly aiming at, refactoring this in such a way to make that swap over easier to manage.

ruslandoga:
Definitely! I actually also decided to mostly go with exceptions everywhere instead of error tuples :) 

I think I'll close this PR again (for now) and continue my experiments in https://github.com/ruslandoga/xqlite. I think I'm slowly discovering an API that is quite flexible for a NIF-only library :)

warmwaffles:
sounds good.

warmwaffles:
That split out for the bind seems like a reasonable thing to do. Do you think we can backport that bit in?

ruslandoga:
Regarding https://github.com/elixir-ecto/ecto_sql/issues/638, with "individual binds" we can do something like this without much performance penalty.

```elixir
def bind_integer(stmt, idx, i) when i < 2147483647 and i > -2147483648 do
  bind_int_nif(stmt, idx, i)
end

def bind_integer(stmt, idx, i) do
  bind_int64_nif(stmt, idx, i)
end
```

warmwaffles:
I personally think probably just using int64 for all integers is the safest thing. I don't know if there is a real use case to _not_ do that.

ruslandoga:
enif_get_int actually seems to be 64bit on 64bit archs as well: https://github.com/erlang/otp/blob/9e5a088ca02b196788260aa8dcd6db3972e6031b/erts/emulator/beam/erl_drv_nif.h#L127-L128

> I don't know if there is a real use case to not do that.

Right, I don't remember when I had to interact with a 32bit system last time. Probably a safe trade-off of performance for cleaner and easier to understand code.

warmwaffles:
I don't think there is a real perf difference checking for it being outside of int32 min and max. That branch logic itself will slow it down. Last I checked with SQLite, it will auto size the integer in the database, but for code binding, you just need to make sure to bind with 64 bit and move on.

Besides, in this case the migration timestamps are integers that will never work for 32 bit integers.

ruslandoga:
> That branch logic itself will slow it down. 

Right: https://github.com/ruslandoga/xqlite/pull/3

PR #274, opened 2024-01-05
Support configuring compile flags using Elixir config

dvic:


warmwaffles (commented):
Oh this is interesting, I didn't think you could do this. I'll check this out and play with it.

dvic:
> Oh this is interesting, I didn't think you could do this. I'll check this out and play with it.

Yeah, I've noticed this option on the docs of https://hexdocs.pm/elixir_make/Mix.Tasks.Compile.ElixirMake.html

We've been using it to change the stack depth (we've run into problems using ecto) and it working great:

```elixir
config :exqlite,
  force_build: true,
  make_env: %{
    "EXQLITE_SYSTEM_CFLAGS" => "-DYYSTACKDEPTH=200",
    "V" => "1"
  }
```

Issue #277, opened 2024-01-30
Possible bug opening a database with an accent in path


jfburdet:
Hi, 
The following code is crashing when the directory containing the database contains accents. I'm not skilled enough to check if the bug is in exqlite code ou kino_db. I apologize in advance if I didn't open the bug in the right project.

```elixir
Mix.install([
  {:kino_db, "~> 0.2.3"},
  {:exqlite, "~> 0.11"}
])

defmodule M do
  def create_and_open_db(path) do
    File.rmdir(path)
    File.mkdir(path)

    db_path = path <> "/database.db"

    System.cmd("sqlite3", [db_path, ".version"])

    opts = [database: db_path]
    {:ok, conn} = Kino.start_child({Exqlite, opts})

    Exqlite.query!(conn, "PRAGMA table_list", []) |> dbg()
  end
end

# Sucess
M.create_and_open_db("./no_accent")

# Failure
M.create_and_open_db("./répertoire_accentué")
``

warmwaffles:
Oh this is interesting, I'll need to dig into it a little bit more. This is probably a mistake in handling the binary coming from erlang into the nif when calling `sqlite3_open`.

PR #278, opened 2024-02-16
Pass binary filename to sqlite3_open

greg-rychlewski:
Closes https://github.com/elixir-sqlite/exqlite/issues/277

This looks like it's your issue since Elixir strings at UTF-8 encoded. But I have to admit I'm not sure why it was latin-1 before. So I might be stepping on a landmine without knowing it.


greg-rychlewski:
Ok I learned very quickly why it was `ERL_NIF_LATIN1` before. It's because `ERL_NIF-UTF8` is new. So instead of passing it as a charlist I pass it as a binary and then convert it to a proper null terminated string inside the nif.

greg-rychlewski:
i'm pretty sure the failures are race conditions because they were not touched in this pr

greg-rychlewski:
I guess it's not a race condition =(

warmwaffles:
I'll take a look at this later in the evening.

warmwaffles:
This _may_ be an issue with how the filesystem represents utf8 characters in windows. I'm not a windows dev, but I think we'll need to dig in to wchar stuff for the filesystem. I suspect it's utf16.

warmwaffles:
@greg-rychlewski it's been a while, can you try re-running the CI. I can't see where to kick it off on my end.

greg-rychlewski:
@warmwaffles done

PR #279, opened 2024-03-10
Use minimum NIF version for each target

cocoa-xu:
Hi this PR updated `elixir_make` to `v0.8.2` so that we can now compile only once for each target with the corresponding minimum NIF version instead of compiling one binary for each pair of target and NIF version.


warmwaffles (approved):
Thank you!

PR #281, opened 2024-03-23
Update sqlite to v3.45.2

elepedus:
Updating SQLite to v3.45.2 to gain improved JSON(B) support.
https://www.sqlite.org/releaselog/3_45_1.html


warmwaffles:
Thank you, I've been meaning to bump the version lately.

PR #284, opened 2024-04-08
Accept a list of modes to open db with nomutex option.

ColaCheng:
In our use case, we only read the database and never write the data. We found this flag can open db with no mutex. It should be safe to open db as read-only and no mutex.

Ref: https://www.sqlite.org/c3ref/open.html

### DIFF
--- a/lib/exqlite/connection.ex
+++ b/lib/exqlite/connection.ex
@@ -58,6 +58,7 @@ defmodule Exqlite.Connection do
 
   @type connection_opt() ::
           {:database, String.t()}
+          | {:mode, :readwrite | :readonly | :readonly_nomutex}
           | {:journal_mode, journal_mode()}
           | {:temp_store, temp_store()}
           | {:synchronous, synchronous()}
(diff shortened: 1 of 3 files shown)


warmwaffles:
@ColaCheng I haven't messed with readonly no mutex before. Why wouldn't this be the default for readonly?

ColaCheng:
> @ColaCheng I haven't messed with readonly no mutex before. Why wouldn't this be the default for readonly?

Thank you for review. I made readonly along with no mutex by default.

warmwaffles:
I'll need to look at the documentation to figure out if no mutex should be the default. We are compiling sqlite with thread safety enabled, so this flag makes sense. But if we disabled thread safety on the compilation this would effectively be a no-op.

I'll accept this once I learn more

warmwaffles:
@ColaCheng I think we'll need to modify mode to allow a list of modes added. After toying with this some, it's probably best to let the caller explicitly state the modes they want.

I'm thinking something like

```
mode: :readonly
mode: :readwrite
mode: [:readonly, :nomutex]
```

If the mode is just `:nomutex` we should raise an exception to the programmer to specify `:readonly` or `:readwrite`. Because technically, `:readwrite` is allowed with `:nomutex` although not recommended.

ColaCheng:
Hi @warmwaffles, updated according to the comments. Please review it. Thanks!

warmwaffles (approved):

PR #286, opened 2024-05-06
Update mvsqlite

fire:


fire:
Sorry for pushing to the wrong repo.

warmwaffles:
@fire nice. Are you maintaining a parallel fork but with an mvsqlite implementation?

fire:
Yeah! I'm not confident in my elixir-mvsqlite port, so I've been keeping it parallel since clustering sqlite is an obscure art that 3-6 companies in the world have done but is still arcane.

PR #294, opened 2024-09-14
Attempt to stop windows flappy CI tests

warmwaffles:
Trying to figure out how to stop these tests from being flappy. If they continue to flap I may drop support all together.


warmwaffles:
Alas I can not stop the flap.

schutm:
That was very quick. Works perfectly now.

PR #298, opened 2024-10-09
Add individual binds

ruslandoga:
This PR adds ability to set [individual binds](https://github.com/elixir-sqlite/exqlite/pull/273#issuecomment-2400122328) and re-implements `bind/3` using the new `bind_*` functions. And also deprecates `bind/3` in favor of `bind/2` since we don't need conn reference anymore.


warmwaffles:
> Should `Sqlite3.bind` be rewritten using them?

Absolutely, this would mean we can toss my old implementation.

lib/exqlite/sqlite3.ex
```diff
@@ -303,6 +303,94 @@ defmodule Exqlite.Sqlite3 do
...
+      iex> {:ok, conn} = Sqlite3.open(":memory:", [:readonly])
+      iex> {:ok, stmt} = Sqlite3.prepare(conn, "SELECT ?")
+      iex> Sqlite3.bind_integer(stmt, 1, 42)
+      :ok
+
+  """
+  @spec bind_integer(statement, non_neg_integer, integer) :: :ok
+  def bind_integer(stmt, index, integer) do
+    case Sqlite3NIF.bind_integer(stmt, index, integer) do
+      0 = _SQLITE_OK -> :ok
```
warmwaffles, line 352: Linter doesn't like this 😢 besides doing something like `^@SQLITE_OK -> :ok` I think you can simply downcase it here.

c_src/sqlite3_nif.c
```diff
@@ -638,6 +638,135 @@ exqlite_bind(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     return make_atom(env, "ok");
 }
 
+static ERL_NIF_TERM
+exqlite_bind_text(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
+{
+    assert(argc == 3);
```
warmwaffles, line 644: When this is compiled in release mode, these asserts disappear. I think we need to do a check and raise immediately.

```c
if (argc != 3) {
  ERL_NIF_TERM badarg = make_message(env, "invalid arguments");
  return enif_raise_exception(env, badarg);
}
```

ruslandoga: I think it's OK. We only define /3 function.

warmwaffles: True. Just my crazy side of not wanting to crash the erlang VM because someone got wise and invoked it as a `/2` somehow.

warmwaffles: Yea just ignore me here, I can't think of a way to invoke this from erlang or elixir in a manner that gives us too many args or not enough.

lib/exqlite/sqlite3.ex
ruslandoga: https://github.com/elixir-sqlite/exqlite/pull/298/commits/895de04a7041e32731367def3b2f5ec829fc59eb

ruslandoga:
Credo doesn't like me. But that's OK! I don't like it either.

lib/exqlite/connection.ex
```diff
@@ -652,12 +652,12 @@ defmodule Exqlite.Connection do
...
-    case Sqlite3.bind(state.db, ref, params) do
-      :ok ->
-        {:ok, query}
-
-      {:error, reason} ->
-        {:error, %Error{message: to_string(reason), statement: statement}, state}
+    try do
+      Sqlite3.bind(ref, params)
+    rescue
+      e -> {:error, %Error{message: Exception.message(e), statement: statement}, state}
```
ruslandoga, line 658: We lose useful stacktrace here (which `bind_*` was called? helps identify which type was expected and what value was received), but this can probably be handled later.

lib/exqlite/sqlite3.ex
```diff
@@ -110,25 +110,91 @@ defmodule Exqlite.Sqlite3 do
...
+          | Date.t()
+          | Time.t()
+          | number
+          | iodata
+          | {:blob, iodata}
+          | atom
+
+  @deprecated "Use `bind/2` instead"
+  @spec bind(db, statement, [bind_value]) :: :ok
+  def bind(_conn, stmt, args), do: bind(stmt, args)
```
ruslandoga, line 137: I think all functions that receive both db and stmt can be updated to only require stmt since it has all the necessary info and we can use `sqlite3_db_handle(stmt)` to get the DB if needed.

lib/exqlite/sqlite3.ex
```diff
@@ -110,25 +110,91 @@ defmodule Exqlite.Sqlite3 do
...
+      i when is_integer(i) -> bind_integer(stmt, idx, i)
+      f when is_float(f) -> bind_float(stmt, idx, f)
+      b when is_binary(b) -> bind_text(stmt, idx, b)
+      b when is_list(b) -> bind_text(stmt, idx, IO.iodata_to_binary(b))
+      nil -> bind_null(stmt, idx)
+      :undefined -> bind_null(stmt, idx)
+      a when is_atom(a) -> bind_text(stmt, idx, Atom.to_string(a))
+      {:blob, b} when is_binary(b) -> bind_blob(stmt, idx, b)
+      {:blob, b} when is_list(b) -> bind_blob(stmt, idx, IO.iodata_to_binary(b))
+      _other -> raise ArgumentError, "unsupported type: #{inspect(param)}"
```
ruslandoga, line 190: I removed `BindError` to standardize on `ArgumentError`, since it was already used in `convert`.

c_src/sqlite3_nif.c
```diff
@@ -448,194 +448,124 @@ exqlite_prepare(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
 
-    parameter_count = (unsigned int)sqlite3_bind_parameter_count(statement->statement);
-    if (parameter_count != argument_list_length) {
-        return make_error_tuple(env, "arguments_wrong_length");
-    }
+    statement_t* statement;
+    if (!enif_get_resource(env, argv[0], statement_type, (void**)&statement))
+        return raise_badarg(env, argv[0]);
 
-    sqlite3_reset(statement->statement);
```
ruslandoga, line 620: Forgot about this one.

c_src/sqlite3_nif.c
```diff
@@ -1282,7 +1244,13 @@ static ErlNifFunc nif_funcs[] = {
   {"execute", 2, exqlite_execute, ERL_NIF_DIRTY_JOB_IO_BOUND},
   {"changes", 1, exqlite_changes, ERL_NIF_DIRTY_JOB_IO_BOUND},
   {"prepare", 2, exqlite_prepare, ERL_NIF_DIRTY_JOB_IO_BOUND},
-  {"bind", 3, exqlite_bind, ERL_NIF_DIRTY_JOB_IO_BOUND},
+  {"reset", 1, exqlite_reset, ERL_NIF_DIRTY_JOB_CPU_BOUND},
```
ruslandoga, line 1247: I'm using dirty CPU scheduler here because of the mutex in https://github.com/sqlite/sqlite/blob/7f5a10e4ba20826b03ea898945f3ae9138b5568e/src/vdbeapi.c#L136

c_src/sqlite3_nif.c
ruslandoga: https://github.com/elixir-sqlite/exqlite/pull/298/commits/c384bd17f6613aaf1f7a45dba5432a84347e099a

lib/exqlite/connection.ex
warmwaffles: Can we reraise with a reformed error and retain the stack trace?

lib/exqlite/sqlite3.ex
warmwaffles: I added `BindError` because I could not figure out how to raise it directly from the NIF and I wanted to pass the term that had failed. I'm not attached to it.

> Maybe it could happen if blob or text are too large?

AFAIK this probably isn't possible under normal circumstances. You are allowed to redefine what the max limit of blobs can be when compiling, so you could in theory test this yourself locally by defining `-DSQLITE_MAX_LENGTH=4096` and doing a quick test to see the behavior.

https://www.sqlite.org/limits.html

lib/exqlite/connection.ex
ruslandoga: Maybe. I don't have much context on this function or how DBConnection handles exceptions from within .... I think it's safer for now to keep the previous behavior of not raising and returning an error triplet. I tried to think of various scenarios in https://github.com/elixir-sqlite/exqlite/pull/298#discussion_r1792817319 and right now I don't see anything useful being lost.

lib/exqlite/sqlite3.ex
ruslandoga: > raise it directly from the NIF 

AFAIK there is no way to do this right now, but there were some plans to add `error_info` to erl_nif APIs. `error_info` is already used in bifs and get's picked up by Elixir as well: https://github.com/elixir-lang/elixir/blob/be84111523cc18498b68949e4c2058eacb44fa00/lib/elixir/lib/exception.ex#L2483 I think that's how ets started returning not just "badarg" but

```elixir
iex> :ets.new "asdf", []
** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an atom

    (stdlib 6.0) :ets.new("asdf", [])
    iex:1: (file)
```

c_src/sqlite3_nif.c
warmwaffles: Yea this is fine

warmwaffles:
Yea this looks good. Let's get this merged and then iterate one this.

PR #299, opened 2024-10-09
Pre-define atoms

ruslandoga:
This PR predefines atoms like Rustler does. I think this improves readability and kind of makes it safer against typos. Once or if #298 is merged, many of the currently used atoms could be replaced with `raise_badarg()` function. I'll also leave some notes on atoms that I think shouldn't be atoms and rather SQLite errors.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -10,7 +10,36 @@
 #include <erl_nif.h>
 #include <sqlite3.h>
 
-#define MAX_ATOM_LENGTH 255
+static ERL_NIF_TERM am_ok;
+static ERL_NIF_TERM am_error;
+static ERL_NIF_TERM am_badarg;
+static ERL_NIF_TERM am_nil;
+static ERL_NIF_TERM am_out_of_memory;
+static ERL_NIF_TERM am_done;
+static ERL_NIF_TERM am_row;
+static ERL_NIF_TERM am_rows;
+static ERL_NIF_TERM am_invalid_filename;
+static ERL_NIF_TERM am_invalid_flags;
+static ERL_NIF_TERM am_database_open_failed;
+static ERL_NIF_TERM am_failed_to_create_mutex;
+static ERL_NIF_TERM am_invalid_connection;
+static ERL_NIF_TERM am_sql_not_iolist;
+static ERL_NIF_TERM am_connection_closed;
+static ERL_NIF_TERM am_invalid_statement;
+static ERL_NIF_TERM am_invalid_chunk_size;
+static ERL_NIF_TERM am_busy;
+static ERL_NIF_TERM am_invalid_column_count;
+static ERL_NIF_TERM am_transaction;
+static ERL_NIF_TERM am_idle;
+static ERL_NIF_TERM am_database_name_not_iolist;
+static ERL_NIF_TERM am_serialization_failed;
+static ERL_NIF_TERM am_deserialization_failed;
+static ERL_NIF_TERM am_invalid_enable_load_extension_value;
+static ERL_NIF_TERM am_insert;
+static ERL_NIF_TERM am_delete;
+static ERL_NIF_TERM am_update;
+static ERL_NIF_TERM am_invalid_pid;
+static ERL_NIF_TERM am_log;
 
 static ErlNifResourceType* connection_type       = NULL;
 static ErlNifResourceType* statement_type        = NULL;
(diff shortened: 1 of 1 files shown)


c_src/sqlite3_nif.c
```diff
@@ -12,6 +12,36 @@
...
+static ERL_NIF_TERM am_ok;
+static ERL_NIF_TERM am_error;
+static ERL_NIF_TERM am_nil;
+static ERL_NIF_TERM am_out_of_memory;
+static ERL_NIF_TERM am_done;
+static ERL_NIF_TERM am_row;
+static ERL_NIF_TERM am_rows;
+static ERL_NIF_TERM am_invalid_filename;
+static ERL_NIF_TERM am_invalid_flags;
+static ERL_NIF_TERM am_database_open_failed;
```
ruslandoga, line 24: This should be rc + exception, since otherwise it's hard to understand why the open failed. Could be user permissions, could be something else. I think I actually had a problem with this one in the past.

c_src/sqlite3_nif.c
```diff
@@ -12,6 +12,36 @@
...
+static ERL_NIF_TERM am_failed_to_create_mutex;
+static ERL_NIF_TERM am_invalid_connection;
+static ERL_NIF_TERM am_sql_not_iolist;
+static ERL_NIF_TERM am_connection_closed;
+static ERL_NIF_TERM am_invalid_statement;
+static ERL_NIF_TERM am_invalid_chunk_size;
+static ERL_NIF_TERM am_busy;
```
ruslandoga, line 31: This should be rc + exception, since the error message might contain useful info.

c_src/sqlite3_nif.c
```diff
@@ -12,6 +12,36 @@
...
+static ERL_NIF_TERM am_invalid_column_count;
+static ERL_NIF_TERM am_transaction;
+static ERL_NIF_TERM am_idle;
+static ERL_NIF_TERM am_database_name_not_iolist;
+static ERL_NIF_TERM am_serialization_failed;
+static ERL_NIF_TERM am_deserialization_failed;
```
ruslandoga, line 37: I haven't used this API before, but I think it might also benefit from being rc + exception.

c_src/sqlite3_nif.c
warmwaffles: Oh I saw this being done in the erlang sqlite implementation and didn't understand the optimization being done when I initially implemented this. So I just created the atoms on demand rather than all up front at once.

c_src/sqlite3_nif.c
```diff
@@ -447,197 +437,6 @@ exqlite_prepare(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     return make_ok_tuple(env, result);
 }
 
-static ERL_NIF_TERM
-bind(ErlNifEnv* env, const ERL_NIF_TERM arg, sqlite3_stmt* statement, int index)
```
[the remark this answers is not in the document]
ruslandoga: Yeah, I kind of planned that #298 would get merged and then I would continue this PR but then got distracted :)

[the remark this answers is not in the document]
warmwaffles: That sounds good. Thanks for the help.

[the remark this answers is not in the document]
ruslandoga: I think this PR is done, and I'll open another one to explore replacing some of the current atoms with `raise_badarg()` helper.

c_src/sqlite3_nif.c
```diff
@@ -594,7 +585,7 @@ make_cell(ErlNifEnv* env, sqlite3_stmt* statement, unsigned int i)
               sqlite3_column_bytes(statement, i));
 
         default:
-            return make_atom(env, "unsupported");
+            return am_nil;
```
ruslandoga, line 588: This technically doesn't matter since there are only five types in SQLite.

c_src/sqlite3_nif.c
```diff
@@ -123,62 +154,22 @@ get_sqlite3_error_msg(int rc, sqlite3* db)
...
+make_error_tuple(ErlNifEnv* env, ERL_NIF_TERM reason)
 {
     assert(env);
     assert(reason);
 
-    return enif_make_tuple2(env, make_atom(env, "error"), make_atom(env, reason));
-}
-
-static ERL_NIF_TERM
-make_bind_error(ErlNifEnv* env, ERL_NIF_TERM message, ERL_NIF_TERM argument)
```
ruslandoga, line 160: I forgot to remove this function in the individual binds PR.

c_src/sqlite3_nif.c
warmwaffles: I don't think it is possible to reach this state, I defined it _just in case_ it did happen. I don't know if extensions can even set this to a custom type.

warmwaffles (approved):

PR #300, opened 2024-10-09
Make make_binary faster

ruslandoga:
According to the docs this version would try to allocate small binaries on the process heap, which seems to positively affect [benchmarks](https://github.com/ruslandoga/elixir-sqlite-bench) (xqlite uses this approach).


bench/fetch_all.exs
```diff
@@ -0,0 +1,30 @@
...
+  values(0)
+  union all
+  select i + 1 from cte where i < ?
+)
+select 'hello' || i from cte
+"""
+
+alias Exqlite.Sqlite3
+
+Benchee.run(
```
ruslandoga, line 12: The benchmark will be removed before merging.

c_src/sqlite3_nif.c
```diff
@@ -182,20 +183,12 @@ make_bind_error(ErlNifEnv* env, ERL_NIF_TERM message, ERL_NIF_TERM argument)
...
-        return make_atom(env, "out_of_memory");
-    }
-
-    memcpy(blob.data, bytes, size);
-    term = enif_make_binary(env, &blob);
-    enif_release_binary(&blob);
-
-    return term;
+    ERL_NIF_TERM bin;
+    const char* data = enif_make_new_binary(env, size, &bin);
```
warmwaffles, line 189: I think I didn't use this function because of the documentation: https://www.erlang.org/doc/apps/erts/erl_nif.html#enif_make_new_binary

> The drawbacks are that the binary cannot be kept between NIF calls and it cannot be reallocated.

Is this saying that we shouldn't store it in a global `static ERL_NIF_TERM` and that it should be safe to pass back to the VM but don't rely on it being present?

ruslandoga: Yes, I think that's what it means. I'm pretty sure it's safe to use since Exqlite only keeps conn and stmt resources between nif calls.

warmwaffles:
Not a problem @ruslandoga thank you so much for the help thus far.

PR #301, opened 2024-10-10
Raise ArgumentError more often

ruslandoga:
Continues #299

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -18,27 +18,17 @@ static ERL_NIF_TERM am_out_of_memory;
 static ERL_NIF_TERM am_done;
 static ERL_NIF_TERM am_row;
 static ERL_NIF_TERM am_rows;
-static ERL_NIF_TERM am_invalid_filename;
-static ERL_NIF_TERM am_invalid_flags;
-static ERL_NIF_TERM am_database_open_failed;
 static ERL_NIF_TERM am_failed_to_create_mutex;
-static ERL_NIF_TERM am_invalid_connection;
-static ERL_NIF_TERM am_sql_not_iolist;
 static ERL_NIF_TERM am_connection_closed;
-static ERL_NIF_TERM am_invalid_statement;
-static ERL_NIF_TERM am_invalid_chunk_size;
 static ERL_NIF_TERM am_busy;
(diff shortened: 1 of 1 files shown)


c_src/sqlite3_nif.c
```diff
@@ -610,7 +572,8 @@ make_row(ErlNifEnv* env, sqlite3_stmt* statement)
 
     columns = enif_alloc(sizeof(ERL_NIF_TERM) * count);
     if (!columns) {
-        return make_error_tuple(env, "out_of_memory");
+        // TODO: should be exception, otherwise it's practically invisible
+        return make_error_tuple(env, am_out_of_memory);
```
ruslandoga, line 576: make_row and maybe other functions might need to be reworked to return this as an error tuple and not an element of a list of rows.

warmwaffles: I agree with this `TODO`, it probably should raise. Although what will most likely happen is that it won't be able to raise because the VM cannot allocate enough memory regardless. The programmer is going to be in for a world of pain.

ruslandoga: Yeah, that's another thing I was thinking about, would any of the current out_of_memory errors every surface or would the app just be OOM-killed. I have seen these errors all SQLite Erlang/Elixir bindings and in https://github.com/jlouis/enacl as well. I think trying to raise is still a good strategy.

warmwaffles: > would any of the current out_of_memory errors ever surface or would the app just be OOM-killed

That's hard to say because I haven't experienced it _yet_. I don't know what happens with the VM when you try to allocate a term and there is no memory available. Looking at `enif_alloc` it returns `NULL` if it fails to allocate.

Initially my target audience for this library was embedded Nerves devices and memory constrained machines running a small elixir application, so it's probably best to figure out a good strategy for handling out of memory. I suspect, probably the best thing to do is to preallocate with the atoms a term or an exception term to return when there is a memory problem. This in theory should not allocate more memory since we would have pre allocated it prior.

ruslandoga: PR: https://github.com/elixir-sqlite/exqlite/pull/304

ruslandoga:
Please see https://github.com/elixir-sqlite/exqlite/pull/300#issuecomment-2602770378

PR #302, opened 2024-10-10
precompile support for `armv7l-linux-gnueabihf`

cocoa-xu:
Long story short: `ErlNapiSInt64` defined in `erl_drv_nif.h` becomes signed 32-bit integer when cross-compiling for an `armv7l` target on an `x86_64` host.

The old but gold printf debugging 🫣

```c
        case SQLITE_INTEGER: {
            int64_t value = sqlite3_column_int64(statement, i);
            sqlite_int64 v = value;
            printf("sqlite3_column_int64: int64_t %lld\r\n", value);
            printf("sqlite3_column_int64: sqlite_int64 %lld\r\n", v);
            printf("sizeof(ErlNapiSInt64): %ld\r\n", sizeof(ErlNapiSInt64));
            return enif_make_int64(env, v);
        }
```

```
sqlite3_column_int64: int64_t 20221010123504
sqlite3_column_int64: sqlite_int64 20221010123504
sizeof(ErlNapiSInt64): 4
```

So the potential solution is to compile it in an `armv7l-linux-gnueabihf` docker container.

Relevant issue: https://github.com/elixir-ecto/ecto_sql/issues/638

### DIFF
--- a/.github/workflows/linux-precompile.yml
+++ b/.github/workflows/linux-precompile.yml
@@ -9,6 +9,59 @@ permissions:
   contents: write
 
 jobs:
+  armv7l-linux-gnueabihf:
+    runs-on: ubuntu-latest
+    env:
+      ARCH: armv7l
+      TARGET: armv7l-linux-gnueabihf
+    strategy:
+      matrix:
+        job:
+          - {otp: "27.1.1", elixir: "1.17"}
+          - {otp: "25.3.2.14", elixir: "1.16"}
+
+    name: Linux armv7l - OTP ${{ matrix.job.otp }} - Elixir ${{ matrix.job.elixir }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


cocoa-xu:
Actually, I might have a better solution

cocoa-xu:
Now it should be much simpler, and it turns out we have to use the header files from the armv7l version of OTP.

warmwaffles:
Awesome. It will only be a matter of time before someone using an ARM based Windows machine asks for support.

warmwaffles:
Thank you @cocoa-xu I appreciate the investigation.

cocoa-xu:
> Awesome. It will only be a matter of time before someone using an ARM based Windows machine asks for support.
> Thank you @cocoa-xu I appreciate the investigation.

Not a problem! And let me know when we need to do precompile support for ARM based Windows!

Issue #129, opened 2021-03-24
Too many SQL variables error


kevinlang:
While cleaning up the benchmarks in `ecto_sqlite3`, I hit this error due to the benchmarks setup code that used `insert_all` hitting this limit.

Two options I can think of:
1. Simply list this as a limitation in the documentation
2. Have Exqlite detect and avoid this case via intelligent batching 

FWIW you can see the code that hits this limit by looking at `sqlite.c:101742`

warmwaffles:
I don't know what the performance implications of bumping that limit up is. We should go look at sqlite3 in ruby to see what it does for active record.

I'd err towards the side of documentation versus bumping the limit.

warmwaffles:
@kevinlang maybe we could figure out a way to support it just being left at default but also allowing someone who knows what they are doing to bump the limit up.

thomastay:
Ran into this today. The limit is well documented in point 9 here: https://sqlite.org/limits.html, but exqlite seems to hit this error below the 32k variable limit somehow. There's also supposedly a way to reduce this limit at runtime, which might complicate how we might want to expose this.

warmwaffles:
@thomastay curious what were you doing to hit that limit? bulk inserting?

PR #306, opened 2024-10-14
Add insert_all

ruslandoga:
This PR adds a helper NIF that can be used to bulk-insert "unlimited" number of rows.

Should be enough to clоse https://github.com/elixir-sqlite/exqlite/issues/129 once it gets into `ecto_sqlite3`

Regarding [performance:](https://github.com/elixir-sqlite/exqlite/issues/129#issuecomment-937400559) XQLite uses this approach in `insert_all` benchmark in https://github.com/ruslandoga/elixir-sqlite-bench

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -695,8 +695,9 @@ static ERL_NIF_TERM
 exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 {
     statement_t* statement;
-    if (!enif_get_resource(env, argv[0], statement_type, (void**)&statement))
+    if (!enif_get_resource(env, argv[0], statement_type, (void**)&statement)) {
         return raise_badarg(env, argv[0]);
+    }
 
     int stmt_param_count = sqlite3_bind_parameter_count(statement->statement);
     int types_array[stmt_param_count];
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -728,8 +728,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         for (unsigned int i = 1; i <= stmt_param_count; i++) {
             ERL_NIF_TERM param;
 
-            if (!enif_get_list_cell(env, head, &param, &head))
+            if (!enif_get_list_cell(env, head, &param, &head)) {
                 return raise_badarg(env, param);
+            }
 
             if (enif_is_identical(param, am_nil)) {
                 rc = sqlite3_bind_null(statement->statement, i);
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -785,8 +785,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         }
 
         rc = sqlite3_step(statement->statement);
-        if (rc != SQLITE_DONE)
+        if (rc != SQLITE_DONE) {
             return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
+        }
 
         rows = tail;
     }
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -738,8 +738,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
                 switch (types_array[i - 1]) {
                     case SQLITE_INTEGER: {
                         ErlNifSInt64 i64;
-                        if (!enif_get_int64(env, param, &i64))
+                        if (!enif_get_int64(env, param, &i64)) {
                             return raise_badarg(env, param);
+                        }
 
                         rc = sqlite3_bind_int64(statement->statement, i, i64);
                         break;
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -728,7 +728,7 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         sqlite3_reset(statement->statement);
 
         // bind row
-        for (unsigned int i = 1; i <= stmt_param_count; i++) {
+        for (int i = 1; i <= stmt_param_count; i++) {
             ERL_NIF_TERM param;
 
             if (!enif_get_list_cell(env, head, &param, &head)) {
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -779,8 +779,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
                 }
             }
 
-            if (rc != SQLITE_OK)
+            if (rc != SQLITE_OK) {
                 return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
+            }
         }
 
         rc = sqlite3_step(statement->statement);
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -713,8 +713,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
         }
 
         int type;
-        if (!enif_get_int(env, head, &type))
+        if (!enif_get_int(env, head, &type)) {
             return raise_badarg(env, types);
+        }
 
         types_array[i] = type;
         types          = tail;
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -758,8 +758,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 
                     case SQLITE_TEXT: {
                         ErlNifBinary text;
-                        if (!enif_inspect_binary(env, param, &text))
+                        if (!enif_inspect_binary(env, param, &text)) {
                             return raise_badarg(env, param);
+                        }
 
                         rc = sqlite3_bind_text(statement->statement, i, (char*)text.data, text.size, SQLITE_TRANSIENT);
                         break;
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -748,8 +748,9 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 
                     case SQLITE_FLOAT: {
                         double f64;
-                        if (!enif_get_double(env, param, &f64))
+                        if (!enif_get_double(env, param, &f64)) {
                             return raise_badarg(env, param);
+                        }
 
                         rc = sqlite3_bind_double(statement->statement, i, f64);
                         break;
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -692,6 +692,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 }
 
 static ERL_NIF_TERM
+exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
+{
+    statement_t* statement;
+    if (!enif_get_resource(env, argv[0], statement_type, (void**)&statement))
+        return raise_badarg(env, argv[0]);
+
+    int stmt_param_count = (unsigned int)sqlite3_bind_parameter_count(statement->statement);
+    int types_array[stmt_param_count];
+
+    ERL_NIF_TERM types = argv[1];
+    ERL_NIF_TERM rows  = argv[2];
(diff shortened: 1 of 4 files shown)


c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    // process rows
+    while (enif_get_list_cell(env, rows, &head, &tail)) {
+        sqlite3_reset(statement->statement);
+
+        // bind row
+        for (unsigned int i = 1; i <= stmt_param_count; i++) {
+            ERL_NIF_TERM param;
+
+            if (!enif_get_list_cell(env, head, &param, &head))
+                return raise_badarg(env, param);
```
warmwaffles, line 732:
```suggestion
            if (!enif_get_list_cell(env, head, &param, &head)) {
                return raise_badarg(env, param);
            }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+            if (enif_is_identical(param, am_nil)) {
+                rc = sqlite3_bind_null(statement->statement, i);
+            } else {
+                switch (types_array[i - 1]) {
+                    case SQLITE_INTEGER: {
+                        ErlNifSInt64 i64;
+                        if (!enif_get_int64(env, param, &i64))
+                            return raise_badarg(env, param);
```
warmwaffles, line 741:
```suggestion
                        if (!enif_get_int64(env, param, &i64)) {
                            return raise_badarg(env, param);
                        }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+                        rc = sqlite3_bind_int64(statement->statement, i, i64);
+                        break;
+                    }
+
+                    case SQLITE_FLOAT: {
+                        double f64;
+                        if (!enif_get_double(env, param, &f64))
+                            return raise_badarg(env, param);
```
warmwaffles, line 750:
```suggestion
                        if (!enif_get_double(env, param, &f64)) {
                            return raise_badarg(env, param);
                        }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+                        rc = sqlite3_bind_double(statement->statement, i, f64);
+                        break;
+                    }
+
+                    case SQLITE_TEXT: {
+                        ErlNifBinary text;
+                        if (!enif_inspect_binary(env, param, &text))
+                            return raise_badarg(env, param);
```
warmwaffles, line 759:
```suggestion
                        if (!enif_inspect_binary(env, param, &text)) {
                            return raise_badarg(env, param);
                        }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+                        rc = sqlite3_bind_text(statement->statement, i, (char*)text.data, text.size, SQLITE_TRANSIENT);
+                        break;
+                    }
+
+                    case SQLITE_BLOB: {
+                        ErlNifBinary blob;
+                        if (!enif_inspect_binary(env, param, &blob))
+                            return raise_badarg(env, param);
```
warmwaffles, line 768:
```suggestion
                        if (!enif_inspect_binary(env, param, &blob)) {
                            return raise_badarg(env, param);
                        }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+                        rc = sqlite3_bind_blob(statement->statement, i, (char*)blob.data, blob.size, SQLITE_TRANSIENT);
+                        break;
+                    }
+                }
+            }
+
+            if (rc != SQLITE_OK)
+                return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
```
warmwaffles, line 777:
```suggestion
            if (rc != SQLITE_OK) {
                return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
            }
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,102 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    int types_array[stmt_param_count];
+
+    ERL_NIF_TERM types = argv[1];
+    ERL_NIF_TERM rows  = argv[2];
+    ERL_NIF_TERM head, tail;
+
+    // process types
+    for (unsigned int i = 0; i < stmt_param_count; i++) {
+        if (!enif_get_list_cell(env, types, &head, &tail))
+            return raise_badarg(env, types);
```
warmwaffles, line 711:
```suggestion
        if (!enif_get_list_cell(env, types, &head, &tail)) {
            return raise_badarg(env, types);
        }
```

lib/exqlite/sqlite3.ex
```diff
@@ -277,6 +277,42 @@ defmodule Exqlite.Sqlite3 do
     fetch_all(conn, statement, chunk_size)
   end
 
+  @doc """
+  Bulk-inserts rows into a prepared statement. Must be called inside a transaction.
+
```
warmwaffles, line 282:
```suggestion
  @doc """
  Bulk-inserts rows into a prepared statement. Must be called inside a transaction.

  The `types` parameter is type of each column in the order they are specified 
  in the parameter `rows`. For example, `[["a", 1, 1.0]]` would have `types`
  specified as `[:text, :integer, :float]`.
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,112 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+    int stmt_param_count = sqlite3_bind_parameter_count(statement->statement);
+    int types_array[stmt_param_count];
+
+    ERL_NIF_TERM types = argv[1];
+    ERL_NIF_TERM rows  = argv[2];
+    ERL_NIF_TERM head, tail;
+
+    // process types
+    for (unsigned int i = 0; i < stmt_param_count; i++) {
```
ruslandoga, line 710:
```suggestion
    for (int i = 0; i < stmt_param_count; i++) {
```

lib/exqlite/sqlite3.ex
```diff
@@ -277,6 +277,45 @@ defmodule Exqlite.Sqlite3 do
...
+  The `types` parameter is type of each column in the order they are specified 
+  in the parameter `rows`. For example, `[["a", 1, 1.0]]` would have `types`
+  specified as `[:text, :integer, :float]`.
+      iex> {:ok, conn} = Sqlite3.open(":memory:", [:readonly])
```
ruslandoga, line 286:
```suggestion

      iex> {:ok, conn} = Sqlite3.open(":memory:", [:readonly])
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,112 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    }
+
+    int rc;
+
+    // process rows
+    while (enif_get_list_cell(env, rows, &head, &tail)) {
+        sqlite3_reset(statement->statement);
+
+        // bind row
+        for (unsigned int i = 1; i <= stmt_param_count; i++) {
```
warmwaffles, line 731:
```suggestion
        for (int i = 1; i <= stmt_param_count; i++) {
```

warmwaffles:
This looks good.

warmwaffles:
I do not like how often the windows test flaps.

warmwaffles:
Ugh

```
c_src\sqlite3_nif.c(703): error C2057: expected constant expression
c_src\sqlite3_nif.c(703): error C2466: cannot allocate an array of constant size 0
c_src\sqlite3_nif.c(703): error C2133: 'types_array': unknown size
```

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,112 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+static ERL_NIF_TERM
+exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
+{
+    statement_t* statement;
+    if (!enif_get_resource(env, argv[0], statement_type, (void**)&statement)) {
+        return raise_badarg(env, argv[0]);
+    }
+
+    int stmt_param_count = sqlite3_bind_parameter_count(statement->statement);
+    int types_array[stmt_param_count];
```
warmwaffles, line 703: This needs to be dynamically allocated. It won't work with the windows compiler.

ruslandoga: Ah, I see. I'll fix it properly tomorrow!

warmwaffles: I wonder if the Zig compiler would be a better long term option for Windows. 🤔 I'll have to play with that idea later.

warmwaffles: @ruslandoga this is the last bit that we need to update. Otherwise this looks good to me.

ruslandoga: 👋 @warmwaffles 

Right, sorry. Got a bit distracted again. I'll finish it up this weekend!

ruslandoga: I pushed a commit but I don't like it ...

> I wonder if the Zig compiler would be a better long term option for Windows.

Maybe just Clang? It seems to be supported on Windows, and can be use for pre-compilation. We can also document that Windows users, if they want to compile the library from scratch, should use Clang.

It seems like too much effort trying to adapt to different compilers :)

ruslandoga: Side issue: there seems to be something wrong with Makefile, it doesn't recompile on sqlite_nif.c changes.

ruslandoga: Seems to be `cc_precompiler` issue, disabling it makes `mix compile` recompile touched files.

c_src/sqlite3_nif.c
```diff
@@ -700,7 +700,14 @@ exqlite_insert_all(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
     }
 
     int stmt_param_count = sqlite3_bind_parameter_count(statement->statement);
-    int types_array[stmt_param_count];
+
+    // TODO: MSVC can't do VLA
+    // int types_array[stmt_param_count];
+    int* types_array = (int*)malloc(stmt_param_count * sizeof(int));
```
warmwaffles: Use `exqlite_malloc` for this. This way the VM will know about the memory allocated and will count it correctly. Also `exqlite_free` to free the memory.

ruslandoga: Right, I did it in the very next commit: https://github.com/elixir-sqlite/exqlite/pull/306/commits/cf69cd54cec3d111b7e648deac36f028d04ad227

warmwaffles: ah I was walking the changes.

c_src/sqlite3_nif.c
```diff
@@ -691,6 +691,120 @@ exqlite_step(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+            }
+
+            if (rc != SQLITE_OK) {
+                return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
+            }
+        }
+
+        rc = sqlite3_step(statement->statement);
+        if (rc != SQLITE_DONE) {
+            return make_sqlite3_error_tuple(env, rc, sqlite3_db_handle(statement->statement));
```
warmwaffles, line 798: We are missing cleanup code here. We need to `enif_free(types_array)` before all of these error returns.

ruslandoga: Ah, right.

ruslandoga: I think I'll need to find another way to pass the types information. The current approach is getting complicated.

ruslandoga: Sorry!

ruslandoga: > This is just the nature of C.

I think it should be possible to implement `insert_all` or a similar function for bulk inserts with no manual memory allocation. It's only used in a small part of the function -- just to know what types to insert -- but ends up affecting all control flow.

ruslandoga: All `Repo.insert_all` calls would go through this NIF and the performance improvement is quite significant (seems close to x2 from my benchmarks, but potentially higher becuase it allows reusing a single prepared statement for all inserts into a single table) so the I think it's worth it.

PR #310, opened 2024-11-17
force build in dev and test envs

ruslandoga:
Context: https://github.com/elixir-sqlite/exqlite/pull/306#discussion_r1845293400

I wonder if this would also force builds in dev / test downstream? The desired outcome is to only force builds while working on Exqlite itself.


ruslandoga:
Ah, I was using an outdated branch. Seems like the current main already has `-dev` version suffix and it seems to work.

warmwaffles:
I have been debating flipping to something like this and dropping the `-dev` version directive. It causes me to have to commit a change, removing it, tagging it, then another change to add it back in. Quite tedious.

PR #312, opened 2025-01-18
The module_str parameter of open_resource_type has to be NULL

filmor:
As documented in https://www.erlang.org/doc/apps/erts/erl_nif#enif_open_resource_type.

This only leads to problems in debug builds as the value is currently guarded by an assertion.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -990,7 +990,7 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
 
     connection_type = enif_open_resource_type(
       env,
-      "exqlite",
+      NULL,
       "connection_type",
       connection_type_destructor,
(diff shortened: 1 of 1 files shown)


warmwaffles:
Odd. Makes sense as I never build the library in debug mode. Always in `-O3`.

filmor:
This has nothing to do with the library, it's about the BEAM. A debug BEAM will be compiled with the respective assertion and fail to load this NIF, see https://erlangforums.com/t/debug-emulator-fails-to-run-with-exqlite-nif/4399.

warmwaffles:
Makes sense. I haven't built a debug BEAM or used one before. 

@filmor just curious, why were you running a debug build? Just wondering if I should be integrating it into my workflow for NIF development.

filmor:
I wasn't, @alanj853 was :)

PR #317, opened 2025-01-29
Add more detailed error message to database open failures.

dominicletz:


c_src/sqlite3_nif.c
```diff
@@ -319,6 +319,15 @@ exqlite_open(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
 
     rc = sqlite3_open_v2((char*)bin.data, &db, flags, NULL);
     if (rc != SQLITE_OK) {
+        if (db != NULL) {
+            char error_msg[1024];
+            const char* msg = sqlite3_errmsg(db);
+            int code = sqlite3_system_errno(db);
+            snprintf(error_msg, sizeof(error_msg), "%s (errno: %d)", msg, code);
```
warmwaffles, line 326: Let me think about this a little bit more. This changes the method signature from `{:error, atom()}` to `{:error, atom() | String.t()}`. I think maybe the better solution would be to do something like `{:error, {:database_open_failed, code(), String.t()}}`

ruslandoga: I wonder if we can just return `rc` or raise badarg from the nif. The rc can then be wrapped into a proper Elixir return value (or an error) like we did with the binds.

```elixir
def open(path, flags) do
  case open_nif(c_str(path), flags) do
    db when is_reference(db) -> {:ok, db}
    rc -> raise Error, code: rc, message: errstr(rc)
  end
end
```

(taken from https://github.com/ruslandoga/xqlite/blob/master/lib/xqlite.ex#L59-L77 and https://github.com/ruslandoga/xqlite/blob/62cbd0f1441ab62837ea331c6e5447c8892eaa5b/c_src/xqlite_nif.c#L91-L117)

PR #318, opened 2025-02-10
fix bind_* NIF typespecs

ademenev:


ademenev:
```elixir
    now = :erlang.monotonic_time(:millisecond)
    {:ok, stmt} = Exqlite.Sqlite3.prepare(conn, "DELETE FROM queue WHERE expiration <= ?")
    :ok = Exqlite.Sqlite3.bind(stmt, [now])
```
dialyzer complains about the last line:

```
The function call will not succeed.
   
Exqlite.Sqlite3.bind(_stmt :: reference(), [integer(), ...])
   
will never return since the 2nd arguments differ  from the success typing arguments:
(reference(), nil | [])
```

and in exqlite source there is this:

```elixir
  @spec bind_integer(statement, non_neg_integer, integer) :: :ok
  def bind_integer(stmt, index, integer) do
    case Sqlite3NIF.bind_integer(stmt, index, integer) do
      @sqlite_ok -> :ok
      rc -> raise Exqlite.Error, message: errmsg(stmt) || errstr(rc)
    end
  end
```

and this

```elixir
  @spec bind_integer(statement, non_neg_integer, integer) :: :ok
  def bind_integer(_stmt, _index, _integer), do: :erlang.nif_error(:not_loaded)
```

The return value of `Sqlite3NIF.bind_integer` which is typed as :ok  is matched against @sqlite_ok which is 0. That  should always fail, so

```
The pattern can never match the type.
   
 Pattern:
 0
   
 Type:
 :ok
```

warmwaffles:
😮‍💨 I could not figure out the dialyzer errors for the life of me. It didn't occur to me to _actually_ check the return from the nif.

Can you go ahead and add the dialyzer to the CI as well now 😄 

https://github.com/elixir-sqlite/exqlite/blob/fd2492d376b6c3e485e951ade4aec2e7529d185f/.github/workflows/ci.yml#L38

warmwaffles:
I'm almost tempted to turn these integers into atoms in a separate PR. Much easier to decipher WTF.

warmwaffles (approved):

warmwaffles:
I'll chuck the ci check in later. Thanks a ton @ademenev.

ademenev:
> 😮‍💨 I could not figure out the dialyzer errors for the life of me. It didn't occur to me to _actually_ check the return from the nif.
> 
> Can you go ahead and add the dialyzer to the CI as well now 😄
> 
> https://github.com/elixir-sqlite/exqlite/blob/fd2492d376b6c3e485e951ade4aec2e7529d185f/.github/workflows/ci.yml#L38

It actually fails for me, so I guess adding it would just break CI

```
lib/mix/tasks/download_sqlite.ex:1:callback_info_missing
Callback info about the Mix.Task behaviour is not available.
________________________________________________________________________________
Please file a bug in https://github.com/jeremyjh/dialyxir/issues with this message.

Unknown error occurred: %Protocol.UndefinedError{protocol: String.Chars, value: {8, 34}, description: ""}


Legacy warning:
lib/mix/tasks/download_sqlite.ex:8:34: Unknown function 'Elixir.Exqlite.MixProject':sqlite_version/0
________________________________________________________________________________
lib/mix/tasks/test_sqlite_version.ex:1:callback_info_missing
Callback info about the Mix.Task behaviour is not available.
________________________________________________________________________________
lib/mix/tasks/test_sqlite_version.ex:17:unknown_function
Function Mix.raise/1 does not exist.
________________________________________________________________________________
Please file a bug in https://github.com/jeremyjh/dialyxir/issues with this message.

Unknown error occurred: %Protocol.UndefinedError{protocol: String.Chars, value: {9, 38}, description: ""}


Legacy warning:
lib/mix/tasks/test_sqlite_version.ex:9:38: Unknown function 'Elixir.Exqlite.MixProject':sqlite_version/0
________________________________________________________________________________
done (warnings were emitted)
Halting VM with exit status 2
```

Integestingly, ElixirLS does not report those for me, only `mix dialyzer` does


But al least for projects using the library dialyzer does is not complaining

warmwaffles:
I stopped using ElixirLS for the time being. It is slow, eats memory, and causes so much pain. 😭

PR #337, opened 2025-09-25
feat: Add `serialized` as an option to connect.

mpope9:
Thank you for this great library.

This PR adds the `serialized` option to `connect`, which enables this library to deserialize an existing database at startup.

This is especially useful for loading a serialized database while using Ecto.


warmwaffles:
Just make the CI pass. I completely forgot this was a feature we added 4 years ago. How time flies.

PR #343, opened 2026-02-26
Fix deadlock when canceling queries stuck in VDBE or busy handler

mjc:
Partially addresses [elixir-sqlite/exqlite#192](https://github.com/elixir-sqlite/exqlite/issues/192).

## Summary

This PR fixes a deadlock where `DBConnection.disconnect/2` could block forever while closing a connection whose dirty NIF thread was still inside SQLite. The dirty NIF execution model is unchanged; the fix is to make the blocked SQLite operation cancellable so it can release `conn->mutex` and allow close to finish.

The deadlock can happen in two important cases: a long-running statement inside SQLite VDBE execution, or a statement sleeping in SQLite's busy handler while waiting for a lock. In both cases the NIF call holds `conn->mutex`. If the pool times out and disconnect tries to close the connection, close also needs `conn->mutex`, so it waits behind the stuck operation.

## What changed

This adds `Exqlite.Sqlite3.cancel/1`, which sets a connection cancellation flag and calls `sqlite3_interrupt()`. `Exqlite.Connection.disconnect/2` now calls `cancel/1` before `close/1`, so teardown can interrupt a running statement or cause a busy wait to stop retrying.

This replaces SQLite's default busy handler with an Exqlite busy handler. The handler preserves SQLite's default retry delay schedule, stores the timeout in `conn->busy_timeout_ms`, sleeps with `sqlite3_sleep()`, and checks the cancellation flag between sleeps. This avoids the non-cancellable behavior of SQLite's default busy handler while keeping the implementation portable.

This adds `Exqlite.Sqlite3.set_busy_timeout/2` and changes connection setup to use it instead of `PRAGMA busy_timeout`. `PRAGMA busy_timeout` calls `sqlite3_busy_timeout()`, which installs SQLite's default busy handler and replaces any custom handler. The new API updates Exqlite's stored timeout without destroying the cancellable busy handler.

This adds a configurable progress handler interval through `Exqlite.Sqlite3.set_progress_handler_steps/2` and the `:progress_handler_steps` connection option. The default is `1000` VDBE steps. Values less than `1` disable the progress handler for callers that want to avoid that overhead.

The shared cancellation flag is protected by `interrupt_mutex`. The busy handler, progress handler, `cancel/1`, and close/destructor paths read or write that flag under the same lock. The busy-timeout and progress-handler-step fields are also updated under `interrupt_mutex`.

## User-facing behavior

High-level users who configure `:busy_timeout` on `Exqlite.start_link/1` or an Ecto SQLite repo should not need to change anything. The connection setup path now applies that timeout through `Exqlite.Sqlite3.set_busy_timeout/2`.

Low-level users who execute `PRAGMA busy_timeout = ...` directly should migrate to `Exqlite.Sqlite3.set_busy_timeout/2` if they want to keep cancellation of busy waits. Executing the pragma still works as SQLite behavior, but it replaces Exqlite's custom busy handler with SQLite's default handler, so `cancel/1` can no longer stop the busy-handler retry loop before SQLite's busy timeout expires.

`Exqlite.Sqlite3.interrupt/1` remains available for SQLite interruption. `Exqlite.Sqlite3.cancel/1` is the stronger teardown-oriented API because it sets Exqlite's cancellation flag for the busy/progress handlers and also calls `sqlite3_interrupt()`.

## Implementation notes

The busy handler uses SQLite's default delay schedule: `1, 2, 5, 10, 15, 20, 25, 25, 25, 50, 50` ms, then `50` ms for later retries. Cancellation is observed between sleep intervals, so a busy wait exits after the current sleep finishes, with later sleeps capped at 50 ms.

The NIF stores the calling environment and pid while SQLite operations are running so the busy handler can detect when the caller process has died. If the caller is gone, the busy handler sets the cancellation flag and stops retrying.

`cancel/1` and `interrupt/1` avoid taking `conn->mutex` while a query may be running. A running SQLite call holds that mutex, so taking it in the cancellation path would block behind the operation we are trying to cancel. The timeout/progress configuration setters still take `conn->mutex` because they are not used to break a currently running SQLite call.

The NIF resource destructor and close path coordinate through `conn->mutex` and `interrupt_mutex` so connection teardown does not race with cancellation state updates or a concurrent interrupt.

The Mix project now tracks the C source files as external resources, and the Makefile emits dependency files for C headers, so C/NIF changes are rebuilt more reliably.

## Tests

This PR adds regression coverage for busy-timeout behavior, low-level `set_busy_timeout/2`, configurable progress-handler steps, `cancel/1`, connection reuse after cancellation, pool recovery after timeouts, cancellation of busy-handler waits, and close/cancel races.

It also adds sanitizer-tagged stress tests for the cancellation and busy-timeout update paths. Those tests are excluded from the default suite and can be run explicitly with `--include sanitizer`.

### DIFF
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -479,7 +479,7 @@ exqlite_busy_handler(void* arg, int count)
 
     // Wait on the condvar — can be woken early by cancel()
     tw_lock(&conn->cancel_tw);
-    int cancelled = conn->cancelled;
+    cancelled = conn->cancelled;
     if (!cancelled) {
         tw_wait_ms(&conn->cancel_tw, sleep_ms);
         cancelled = conn->cancelled; // snapshot again after waking
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -61,7 +61,7 @@ typedef struct connection
     int authorizer_deny[AUTHORIZER_DENY_SIZE];
 
     // Custom busy handler state
-    volatile int cancelled; // guarded by interrupt_mutex
+    int cancelled; // guarded by interrupt_mutex
     int busy_timeout_ms;
     int progress_handler_steps;
     ErlNifEnv* callback_env; // for enif_is_process_alive
--- a/c_src/sqlite3_nif.c
+++ b/c_src/sqlite3_nif.c
@@ -10,118 +10,6 @@
 #include <erl_nif.h>
 #include <sqlite3.h>
 
-// Platform timed-wait abstraction
-//
-// OTP doesn't provide enif_cond_timedwait, so we wrap pthread_cond_timedwait
-// (POSIX) and SleepConditionVariableCS (Windows). BEAM has two threading
-// backends (pthreads and Win32), so #ifdef _WIN32 covers all platforms.
-
-#ifdef _WIN32
-    #include <windows.h>
-
-typedef struct
-{
-    CRITICAL_SECTION cs;
-    CONDITION_VARIABLE cv;
-} timed_wait_t;
-
-static void
-tw_init(timed_wait_t* tw)
-{
-    InitializeCriticalSection(&tw->cs);
-    InitializeConditionVariable(&tw->cv);
-}
-
-static void
-tw_destroy(timed_wait_t* tw)
-{
-    DeleteCriticalSection(&tw->cs);
-    // Windows CONDITION_VARIABLE has no destroy function
-}
-
-static void
-tw_lock(timed_wait_t* tw)
-{
-    EnterCriticalSection(&tw->cs);
-}
-static void
-tw_unlock(timed_wait_t* tw)
-{
-    LeaveCriticalSection(&tw->cs);
-}
-
-// Returns 0 if signalled, 1 if timed out
-static int
-tw_wait_ms(timed_wait_t* tw, int ms)
-{
-    return SleepConditionVariableCS(&tw->cv, &tw->cs, (DWORD)ms) ? 0 : 1;
-}
-
-static void
-tw_signal(timed_wait_t* tw)
-{
-    WakeConditionVariable(&tw->cv);
-}
-
-#else /* POSIX */
-    #include <pthread.h>
-    #include <time.h>
-    #include <errno.h>
-
-typedef struct
-{
-    pthread_mutex_t mtx;
-    pthread_cond_t cond;
-} timed_wait_t;
-
-static void
-tw_init(timed_wait_t* tw)
-{
-    pthread_mutex_init(&tw->mtx, NULL);
-    // CLOCK_REALTIME: macOS doesn't support CLOCK_MONOTONIC for condvars
-    pthread_cond_init(&tw->cond, NULL);
-}
-
-static void
-tw_destroy(timed_wait_t* tw)
-{
-    pthread_cond_destroy(&tw->cond);
-    pthread_mutex_destroy(&tw->mtx);
-}
-
-static void
-tw_lock(timed_wait_t* tw)
-{
-    pthread_mutex_lock(&tw->mtx);
-}
-static void
-tw_unlock(timed_wait_t* tw)
-{
-    pthread_mutex_unlock(&tw->mtx);
-}
-
-// Returns 0 if signalled, 1 if timed out
-static int
-tw_wait_ms(timed_wait_t* tw, int ms)
-{
-    struct timespec ts;
-    clock_gettime(CLOCK_REALTIME, &ts);
-    ts.tv_sec += ms / 1000;
-    ts.tv_nsec += (ms % 1000) * 1000000L;
-    if (ts.tv_nsec >= 1000000000L) {
-        ts.tv_sec += 1;
-        ts.tv_nsec -= 1000000000L;
-    }
-    return pthread_cond_timedwait(&tw->cond, &tw->mtx, &ts) == ETIMEDOUT ? 1 : 0;
-}
-
-static void
-tw_signal(timed_wait_t* tw)
-{
-    pthread_cond_signal(&tw->cond);
-}
-
-#endif /* _WIN32 */
 
 static ERL_NIF_TERM am_ok;
 static ERL_NIF_TERM am_error;
@@ -174,7 +62,6 @@ typedef struct connection
     int authorizer_deny[AUTHORIZER_DENY_SIZE];
 
     // Custom busy handler state
-    timed_wait_t cancel_tw;
     volatile int cancelled; // volatile for MSVC compat
     int busy_timeout_ms;
     ErlNifEnv* callback_env; // for enif_is_process_alive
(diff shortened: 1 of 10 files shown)


warmwaffles:
This does not fix that issue that is open. There's stuff in that issue that pertains to this, but this does not replace the Dirty NIF execution model.

I'll need to check this out and understand what is going on before we merge it.

test/exqlite/integration_test.exs
```diff
@@ -243,13 +243,18 @@ defmodule Exqlite.IntegrationTest do
     # With the cancellable busy handler (issue #192), disconnect now properly
     # interrupts running queries via the progress handler. So a query that
     # exceeds the checkout timeout may be interrupted rather than completing.
+    # With a 1ms timeout the disconnect can race mid-allocation, so SQLite may
+    # also return SQLITE_NOMEM before the interrupt lands — that's expected here.
```
mjc: sorry about the churn here I forgot I encountered this this morning and then copilot led me astray.

mjc:
If you're open to the polling approach, the busy handler simplifies to roughly:

```c
static int exqlite_busy_handler(void *arg, int count) {
    connection_t *conn = (connection_t *)arg;
    
    if (conn->cancelled) return 0;
    if (!enif_is_process_alive(conn->callback_env, &conn->caller_pid)) return 0;
    
    conn->busy_elapsed_ms += 10;
    if (conn->busy_elapsed_ms >= conn->busy_timeout_ms) return 0;
    
    sqlite3_sleep(10);
    return !conn->cancelled;
}
```

The entire `timed_wait_t` platform abstraction goes away — no condvar, no mutex, no `#ifdef`, no clock concerns. `cancel` just sets `conn->cancelled = 1` and calls `sqlite3_interrupt`. Happy to refactor to this if it works for you.

mjc:
It's not a busy wait — `sqlite3_sleep(10)` calls through to the SQLite VFS `xSleep`, which calls `usleep()`/`nanosleep()` on POSIX and `Sleep()` on Windows. The thread is fully descheduled by the OS and consumes zero CPU while sleeping.

In fact, this is exactly what SQLite's own built-in busy handler ( [sqliteDefaultBusyCallback](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) ) does — it calls `sqlite3OsSleep()` in a loop with increasing delays (1, 2, 5, 10, 15, 20, 25, 50, 100ms). Our polling handler would do the same thing with a fixed 10ms interval, plus a cancellation flag check. The only trade-off vs. a `condvar` is up to 10ms latency before responding to cancellation.

We could even replicate the same delay ramp that sqliteDefaultBusyCallback uses if desired — since that's what `sqlite3_busy_timeout()` already does internally, just without the ability to interrupt.

mjc:
rebased to address a conflict. I could also clean up the history when you're ready to give it another look. but also no rush as I can just keep using my branch

c_src/sqlite3_nif.c
```diff
@@ -292,6 +298,81 @@ statement_release_lock(statement_t* statement)
...
+// so disconnect() wakes within at most one sleep interval (~10ms).
+// ---------------------------------------------------------------------------
+
+static int
+exqlite_busy_handler(void* arg, int count)
+{
+    connection_t* conn = (connection_t*)arg;
+
+    if (conn->cancelled)
+        return 0;
```
warmwaffles: Add missing curly braces `{}`

c_src/sqlite3_nif.c
warmwaffles: When I originally implemented this, I intentionally left off any mutex locking here, but I suspect if `conn->mutex` is set, we should acquire the lock for teardown.

c_src/sqlite3_nif.c
```diff
@@ -1715,6 +1832,65 @@ exqlite_interrupt(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+    if (!enif_get_resource(env, argv[0], connection_type, (void**)&conn)) {
+        return make_error_tuple(env, am_invalid_connection);
+    }
+
+    if (!enif_get_int(env, argv[1], &timeout_ms)) {
+        return enif_make_badarg(env);
+    }
+
+    conn->busy_timeout_ms = timeout_ms;
```
warmwaffles, line 1937: This is missing an acquire for a lock on `mutex`

c_src/sqlite3_nif.c
```diff
@@ -1715,6 +1832,65 @@ exqlite_interrupt(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+
+    if (argc != 1) {
+        return enif_make_badarg(env);
+    }
+
+    if (!enif_get_resource(env, argv[0], connection_type, (void**)&conn)) {
+        return make_error_tuple(env, am_invalid_connection);
+    }
+
+    conn->cancelled = 1;
```
warmwaffles: Similar situation here, we are missing a lock on the conn being manipulated.

c_src/sqlite3_nif.c
```diff
@@ -292,6 +298,81 @@ statement_release_lock(statement_t* statement)
...
+// Must be called while holding conn->mutex.
+static inline void
+connection_stash_caller(connection_t* conn, ErlNifEnv* env)
+{
+    conn->callback_env = env;
+    enif_self(env, &conn->caller_pid);
+    conn->cancelled = 0;
+}
+
+// Clear the stashed caller after a db operation completes.
```
warmwaffles, line 420:
```suggestion
// Clear the stashed caller after a db operation completes.
// Assumes that the `conn` has been locked for clearing the
// caller.
```

c_src/sqlite3_nif.c
```diff
@@ -344,14 +423,24 @@ exqlite_open(ErlNifEnv* env, int argc, const ERL_NIF_TERM argv[])
...
+    conn->interrupt_mutex = enif_mutex_create("exqlite:interrupt");
     if (conn->interrupt_mutex == NULL) {
-        // conn->db and conn->mutex are set; the destructor will clean them up.
         enif_release_resource(conn);
         return make_error_tuple(env, am_failed_to_create_mutex);
     }
 
+    // Install our custom busy handler + progress handler
+    sqlite3_busy_handler(db, exqlite_busy_handler, conn);
+    sqlite3_progress_handler(db, 1000, exqlite_progress_handler, conn);
```
warmwaffles: I believe we are going to want `1000` to be customizable. If someone wishes to disable this for perf reasons, a `-1` should be allowed per the documentation. Or if they want a higher number than `1000` depending on their application needs.

.github/workflows/ci.yml
warmwaffles, line 1: revert this

warmwaffles:
I'll be getting to this soon. I've been extremely busy lately.

c_src/sqlite3_nif.c
```diff
@@ -59,6 +59,13 @@ typedef struct connection
     ErlNifMutex* interrupt_mutex;
     ErlNifPid update_hook_pid;
     int authorizer_deny[AUTHORIZER_DENY_SIZE];
+
+    // Custom busy handler state
+    volatile int cancelled; // guarded by interrupt_mutex
```
warmwaffles: I am probably going to merge this and work on it some in another branch.

What is the benefit of throwing `volatile` here versus just having it be a basic `int`? I know that `volatile` is a signal to ensure the compiler doesn't re-arrange order of execution, but what I want to know, is this actually a risk?

mjc: That was an oversight from when I backed out the initial proposed handler. No real benefit now that all reads/writes are guarded by `interrupt_mutex`. `volatile` does not make this thread-safe, and the mutex is what provides the synchronization here, so I removed it.

warmwaffles:
@mjc can you update the PR description with the most accurate information you can? I'm going to proof it and then use it for the commit message.

mjc:
Updated! Happy to help as much as I can, also.

I've been running the previous version (with the condvar) since Feb 1 in a bunch of environments with none of the previous issues and no segfaults.

I still get busy timeouts but that is largely due to not yet quite having exactly the right concurrency model in that specific distributed app - the timeouts make sense and are traced to obvious things I know I need to fix. down like 90% from before though

warmwaffles:
> I still get busy timeouts but that is largely due to not yet quite having exactly the right concurrency model in that specific distributed

I still get this as well, it's been a minor irritant because it backs an MCP server I am running.

warmwaffles:
If you figure it out, don't hesitate to open another PR.

PR #346, opened 2026-05-10
Add Android precompile workflow

### LINKED ISSUE #165: Add arm / android compilation strategy to ci workflow
Would be nice to see this compiled for ARM at minimum and then have an Android compilation

dl-alexandre:
## Summary

- add Android `aarch64-linux-android` and `armv7a-linux-androideabi` targets to the existing `cc_precompiler` configuration
- map Android precompile targets to the Android NDK clang toolchain while keeping the existing Linux/macOS/Windows targets unchanged by default
- add an Android precompile workflow that builds one Android target per matrix job and publishes matching release artifacts on tags

Fixes #165.

## Validation

- `mix format --check-formatted mix.exs`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/android-precompile.yml")'`
- `mix eval --no-compile --no-deps-check 'IO.inspect(CCPrecompiler.all_supported_targets(:fetch))'`
- `git diff --check`
- `mix test`

Note: I validated the normal native build/test path locally on macOS. The Android cross-compile path depends on the GitHub-hosted Linux runner's Android NDK installation, so the new workflow is the first full NDK validation point.

### DIFF
--- a/.github/workflows/android-precompile.yml
+++ b/.github/workflows/android-precompile.yml
@@ -1,0 +1,73 @@
+name: android-precompile
+
+on:
+  push:
+    branches:
+      - "main"
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
+
+jobs:
+  precompile:
+    runs-on: ubuntu-latest
+    env:
+      ImageOS: ubuntu22
+      MIX_ENV: prod
+      ANDROID_API_LEVEL: "24"
+      CC_PRECOMPILER_ONLY_LISTED_TARGETS: "true"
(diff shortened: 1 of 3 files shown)


warmwaffles:
Sorry, I'll get to this soon.

dl-alexandre:
Thanks for the update! No rush at all — just wanted to make sure the PR was still on your radar. Happy to provide any additional context or changes if needed (e.g. for other platforms).

warmwaffles:
Hey sorry I just have been swamped with other things lately. This looks good enough to me. I do need to cut another version soon, we have changes interrupts that I need to get cut.

PR #351, opened 2026-07-14
Close connection after error on open

katafrakt:
As per [documentation](https://www.sqlite.org/c3ref/open.html), even when the database opening failed, we should still run `sqlite3_close()` to free the resources:

> Whether or not an error occurs when it is opened, resources associated with the [database connection](https://www.sqlite.org/c3ref/sqlite3.html) handle should be released by passing it to [sqlite3_close()](https://www.sqlite.org/c3ref/close.html) when it is no longer required.

I verified this against Python implementation, which [does this too](https://github.com/python/cpython/blob/357c6500589ca7e065a6c263accfa1307d93c990/Modules/_sqlite/connection.c#L334-L339). Python also calls `assert` to make sure that closing was clean (it should be per a comment there), but I think it's not a good idea for a NIF to do such thing.

### AI disclosure

The issue was flagged by an AI review while I was working on SQLite backend for Expert. The rest of the work was human.


warmwaffles (approved):
I didn't read the documentation for this:

> Calling sqlite3_close() or sqlite3_close_v2() with a NULL pointer argument is a harmless no-op.

Neat.

I appreciate the disclosure.

PR #352, opened 2026-07-14
Remove duplicated `prepare`

katafrakt:
I started reading `connection.ex` to understand query caching story (before reading in the readme that it's not supported; _hours of reading the code can save minutes of reading the docs_). During that I noticed that there is `prepare` and `prepare_no_cache` that are basically identical and definitely both are not caching. I figures this is some leftover from adding the cache in the past. From #41 I understand it's not coming back, so perhaps it's ok to clean this up.

Also fixes some typespecs that are not in line with the implementation.


warmwaffles:
I wouldn't say it coming back is impossible. I just tossed it at the time because I envisioned access patterns where the queries would vary in composition.

warmwaffles (approved):
This looks reasonable.

Curious, how did you find the incorrect typespecs?

katafrakt:
I'm afraid there's no secret sauce to it, I just tend to sometimes spot such things. I don't claim everything is correct now.

warmwaffles:
Well I appreciate it. I usually can spot those as well, but sometimes callbacks like this slip through because the typespec is terse.

Issue #356, opened 2026-08-21
Custom allocator degrades performance of concurrent reads for complex queries


jearbear:
I'm porting an application from Postgres which has some non trivial queries that take ~20ms to execute. These queries are run concurrently using `Task.async_stream`. When using Postgres, this doesn't really affect the runtime of the individual queries so it yields a nice performance benefit. When I used SQLite (via ecto_sqlite3), I noticed that when run concurrently, they would seem to effectively run serially, like each query will just take longer so the total runtime doesn't improve.

The effect is minimal for simple queries, but much more noticeable for the more involved queries in my app. Which is unfortunate because chains of slow queries are the ones I'm more likely to attempt to speed up via concurrency 😅.

I believe this is happening due to the custom allocators used here: https://github.com/elixir-sqlite/exqlite/blob/main/c_src/sqlite3_nif.c#L1428-L1439. When I disable them, throughput starts scaling nicely with the number of workers as expected. (As an aside, after I set `SQLITE_DEFAULT_MEMSTATUS=0` which yielded another ~x2 speed up)

My thinking is that these queries are complex enough that they result in a lot of allocations and the custom allocator is becoming a serialization bottleneck.

Would you be open to a PR that provides a flag to disable the custom allocator? If I'm understanding the [motivations for its introduction](https://github.com/elixir-sqlite/exqlite/pull/193), I'm personally fine trading off visibility into memory usage for the performance benefits.

If you're not quite convinced this could be the problem, I'm happy to work on building a minimal repro tomorrow. Thanks!

warmwaffles:
I'd be okay with a PR showing a proof of concept yes. What I would like to see is also benchmarks showing the perf penalties of using the custom allocator to be tracked by the VM. My main concern was and still is the tracking of memory consumed by the running process. I did personally have some issues where I had an instance that was limited to 512MB and the process would just OOM but the metrics for BEAM was saying I was fine at around 400MB during heavy load. 

Ultimately the solution was to up the memory limit to 2GB and let it run wild, but also to track the memory consumed by the sqlite process itself.

Anyways, go wild. As long as it's optional by some means that'd be fine with me.

jearbear:
I took some of the queries that are being run in my app to show the range of behavior depending on the complexity of the query.

Here are my results with the default settings (custom allocator left on):
```
accounts x 1: 906.822ms
accounts x 2: 1134.481ms
accounts x 4: 2033.027ms
accounts x 8: 2673.0ms

[...]

PR #357, opened 2026-08-22
Support disabling the Erlang allocator

jearbear:
Allow disabling the Erlang allocator which can sometimes have worse performance characteristics compared to the default one as demonstrated in https://github.com/elixir-sqlite/exqlite/issues/356.

In addition to the added automated tests, I manually tested this against the repro file provided in the issue by setting my fork as a dependency and toggling the `disable_erlang_allocator` setting.


c_src/sqlite3_nif.c
```diff
@@ -1469,6 +1472,17 @@ on_load(ErlNifEnv* env, void** priv, ERL_NIF_TERM info)
...
+    ERL_NIF_TERM disable_erlang_allocator;
+    if (!enif_get_map_value(env, info, enif_make_atom(env, "disable_erlang_allocator"), &disable_erlang_allocator)) {
+        return -1;
+    }
+    erlang_allocator_enabled = enif_is_identical(disable_erlang_allocator, am_false);
+
+    sqlite3_config(SQLITE_CONFIG_GETMALLOC, &default_alloc_methods);
+    if (erlang_allocator_enabled) {
+        sqlite3_config(SQLITE_CONFIG_MALLOC, &methods);
+    }
```
jearbear, line 1484: It seemed simpler to me to unconditionally load and restore the default allocator and only conditionally override the default allocator with the Erlang ones to reduce the amount of branching that needs to happen.

jearbear:
@warmwaffles I'm a little unsure about the testing story here because it doesn't seem like the existing allocator is tested currently.

Adding the `exqlite_erlang_allocator_enabled` NIF seemed like an ok middle ground because I want to at least make sure I'm not accidentally reverting the existing default of using the custom allocator. However, it's hard to test the other case easily because the setting can't be changed without restarting the VM so it would require like spawning another test process at least. I tried it out, but it adds to the test runtime and feels pretty hacky.

warmwaffles:
> I'm a little unsure about the testing story here because it doesn't seem like the existing allocator is tested currently.

IIRC when I implemented it, it was not tested at all. I could not come up with a reasonable way to test it other than just running the library. However, I do think perhaps running the tests twice would be a solution. One runs with the custom allocator and one without it.

c_src/sqlite3_nif.c
warmwaffles: Oh this is far simpler than what I had originally envisioned. I'll take a closer look at this tomorrow. As it stands I don't see a problem with this.

warmwaffles:
Don't worry trying to figure out how to test the allocator being swapped. I'll spin up a separate task where this is configured to be true or false and run it on CI just to make sure nothing insane is being done with either allocator.

warmwaffles (approved):

warmwaffles:
Released under `v0.40.0`

jearbear:
> However, I do think perhaps running the tests twice would be a solution. One runs with the custom allocator and one without it.

This crossed my mind, but I didn't want to be the guy who doubled CI runtime 😅.

Thank you for merging and a generally wonderful library!

